Issue #29, opened 2017-10-30
Decide how to handle proper motion


cwwalter:
Proper motions are currently not handled properly in DM. See:

https://jira.lsstcorp.org/browse/DM-8828

So, I will check on the status of this with the DM team, but if DM does not handle proper motion by the time of DC2 we should consider turning it off. For imSim this would mean setting the proper motion components to zero in the instance catalog.  For PhoSim I believe we would pass the ICRS coordinates into the instance catalog instead of the modified ones when we do the CatSim lookup.  @danielsf and @johnrpeterson can comment.

fjaviersanchez:
I agree that if DM doesn't handle proper motion we should deactivate proper motion. That will make our life easier at the time of matching input and output catalogs.

cwwalter:
@johnrpeterson Can you comment on the above (how we would turn off proper motion in PhoSim instance catalogs)?  Thanks.

cwwalter:
Hi @johnrpeterson;

It is looking like handling  proper motion is not something that is on the DM near term roadmap.  Since we think this caused 20mas astrometric errors for us in DC1 we want to understand if we could turn proper motion off so we can decide what to do.

Is this possible in PhoSim, if so, what would the procedure be to make the instance catalogs?

Thanks.

johnrpeterson:
well this is really at the catalog level.   proper motion is either handled by moving the ra/dec at the appropriate time of the observations.  the other case is using the movingpoint object which is normally used for high proper motion objects that move significantly in seconds (e.g. asteroids/comets).  but for that you would just set it to 0.

is that what you are asking?

john


On Nov 6, 2017, at 9:51 PM, Chris Walter <notifications@github.com<mailto:notifications@github.com>> wrote:


[...]

Issue #17, opened 2017-10-20
Run protoDC2/CatSim catalog through PhoSim


katrinheitmann:
The CatSim output now needs to go through PhoSim to generate the image catalog.

danielsf:
Except for one more lingering question about the meaning of the major and minor axes in the protoDC2 catalog (are they "projected on the sky" or some True three-dimensional quantities), this script

https://github.com/LSSTDESC/gcr-catalogs/blob/master/examples/phosim_descqa.py

does everything you need to produce PhoSim InstanceCatalogs from protoDC2.  I have generated a small catalog and successfully run it through PhoSim.

Is having this example script good enough?

What sort of API do we need?

Who is actually going to be doing the work of generating the images we want for the December sprint week?

rmandelb:
> major and minor axes in the protoDC2 catalog (are they "projected on the sky" or some True 
> three-dimensional quantities)

Two numbers are insufficient to specify the 3D size of an ellipsoid (you would need minor, intermediate, and major axes, not just minor and major).  So I think these must be projected.

danielsf:
Dan Korytov has clarified that inclination is already included in the major/minor axis columns.

That just leaves my other two questions, which really can be summarized in the third: who is going to be generating these images (so that I can talk to that person about how they want the interface to work).

katrinheitmann:
Thanks Scott!! @johnrpeterson suggested that Glenn has time to run it through PhoSim and he is set up to do that at NERSC. John, can you confirm that Glenn is available? Then Glenn and Scott can talk about the interface question. @salmanhabib We should also set it up at Argonne, can you push that forward with John, Glenn, Tom U. etc?

jchiang87:
@danielsf It looks like your `phosim_descqa.py` script only generates objects from the protoDC2 catalog, i.e., does not produce entries for stars, correct?  Also it uses the opsim db file from DC1, i.e., `minion_1016_sqlite_new_dithers.db`.  Is that what we want to use for protoDC2?   I think there are some things to sort out before we hand off things to the people running phosim.

danielsf:
@jchiang87 Yes.  That script was only concerned with getting the protoDC2 objects into the catalogs.

Are we just going to use the catalog-generation infrastructure we used for DC1 (in which case, we can just include the DC2-specific classes defined here to replace the galaxy catalog classes), or are we starting from scratch?

jchiang87:
[...]

PR #45, opened 2017-11-18
bug-fixes and added flag to disable proper motion

jchiang87:


scripts/generateDc2InstCat.py
```diff
@@ -26,6 +29,12 @@ def get_maskedMagNorm(self):
             return raw_norm
         return np.where(raw_norm<self.min_mag, self.min_mag, raw_norm)
 
+    def column_by_name(self, colname):
+        if (not colname.startswith('properMotion')
```
jchiang87: @danielsf Should the radial velocity also be zeroed here, as you suggested in #29 ?

cwwalter:
Since there were a few ways of doing it I just wanted to copy this comment from Andy in the issue over to here:

> Just to expand on @danielsf comments. The proper motions are stored in the
base catalog used for imsim and phosim (i.e. the database). It would be
straightforward to set them (and parallax if wanted) to 0 at the point the
database is queried and it should not change anything else in the
generation of the phosim inputs

Is that how you will do it?

jchiang87:
I've implemented it essentially the way @danielsf recommended via email:  return the `.column_by_name("properMotion[Ra,Dec]")` values as properly sized arrays of zeros.  The only question I have is whether the radial velocity columns also need to be zeroed.  In any case, aside from possibly adding that column to my re-implementation of `column_by_name`, this is done.

jchiang87:
BTW, there's one more fix that needs to be added to this branch:  The protoDC2 galaxies will be truncated at the 5x5 degree boundary as I showed in the figure in #17, but the stars will not, so the phosim feature of omitting sensors without objects will still simulate the unwanted sensors with those stars even though there will be no galaxies on them.   So the same 5x5 boundary needs to be applied to the stars too.

scripts/protoDC2/generateDc2InstCat.py
```diff
@@ -12,20 +13,41 @@
 
 class MaskedPhoSimCatalogPoint(PhoSimCatalogPoint):
 
+    disable_proper_motion = False
+
     min_mag = None
 
     column_outputs = ['prefix', 'uniqueId', 'raPhoSim', 'decPhoSim', 'maskedMagNorm', 'sedFilepath',
-                      'redshift', 'gamma1', 'gamma2', 'kappa', 'raOffset', 'decOffset',
+                      'redshift', 'shear1', 'shear2', 'kappa', 'raOffset', 'decOffset',
```
danielsf: I changed `shear1` and `shear2` to `gamma1` and `gamma2` on master last week.  I can issue a new release this week,since I think both conventions are now floating around.

scripts/protoDC2/protoDC2_valid_obsids.py
```diff
@@ -0,0 +1,66 @@
+import pickle
+import numpy as np
+import pandas as pd
+from lsst.sims.catUtils.utils import ObservationMetaDataGenerator
+
+def fov_overlaps_protoDC2(x, y, half_size=2.5, radius=1.77):
```
danielsf: Should this method us `lsst.sims.utils.angularSeparation` to properly take spherical geometry into account?

(Though I guess it technically doesn't matter for protoDC2, since it is centered at Dec=0)

danielsf:
I do believe that `radialVelocity` should also be set to zero.  All of the methods that calculate proper motion also take `radialVelocity` as an input.

scripts/protoDC2/generateDc2InstCat.py
jchiang87: @danielsf I'm about to make a new branch with this change, the `angularSeparation` fix, and (presumably) a change to make the disabling of the application of proper motion as the default behavior of the `generateDc2InstCat.py` script.   Can you let me know when you've issued the sims release with these changes?

danielsf: Okay.  I will start that today.

Issue #50, opened 2017-12-09
Update generateDc2InstCat.py script to produce object catalogs for imSim


jchiang87:

jchiang87:
It looks like one just needs a subclass of `PhoSimDESCQA` that specifies the outputs (ICRS coordinates in particular) for imSim, i.e., similar to
https://github.com/LSSTDESC/imsim_deep_pipeline/blob/master/python/desc/imsim_deep_pipeline/catsim_utils.py#L46
@danielsf Can you confirm this please?

danielsf:
@jchiang87 

You are correct

PR #51, opened 2017-12-09
enable catalog generation for imsim

jchiang87:
This addresses #50


jchiang87:
Ah, yes, of course.  I'll add that.

jchiang87:
Yes, it will apply those effects...i.e., that part of the code hasn't changed from DC1.

danielsf:
In my opinion, this is ready to merge.

PR #74, opened 2018-01-10
Importing the DC2 Plan back from overleaf

drphilmarshall:
@katrinheitmann Since the DC2 Plan seems fairly stable, let's move it back into this repo where it can live alongside our other documents and scripts. That way we can point directly to it from the issues, and propose changes to it via pull request. I didn't make any changes to the Plan content in this move, I just added the files from the overleaf folder, and updated the README to refer to them. Shall we protect the master branch in the DC2_repo to help us get into good code review habits? I can do this no problem.


katrinheitmann:
Done

Issue #73, opened 2018-01-09
DC2 Cookbook


drphilmarshall:
We have DM pipelines in the form of collections of scripts in various places: for DC2 we will need to pull these together and re-apply to the DC2 images as we make them, starting ASAP. To make sure that we are implementing the data release processing (DRP) pipeline that we want to see, we need a DC2 Cookbook to check against. 

@SimonKrughoff and I discussed this a little, and we decided we should try and write it all down in one DESC Note jupyter notebook. The reasoning is that we might then be able to make it run-nable, which could help with its development. (It's also nice how jupyter makes code blocks easy, you just use a code cell rather than a markdown cell. And GitHub does a nice job of rendering Notebooks so they can be browsed on the DC2 repo web interface.) I can start by copying in the pieces from the Twinkles cookbook ("self-plagiarism is the highest form of self-flattery" - S. Digel). Then we can look through them to check that this is what we want to do. Hope this sounds OK, @katrinheitmann !

I'll assign suitable code reviewers in time, but it'd be great if @jchiang87 @tony-johnson @nicolaschotard and @boutigny could follow along from the beginning, to make sure we are on track. All comments and suggestions welcome!

cwwalter:
We shouldn't forget the repos and scripts that were worked out to do this for DC1. The DIA is of course new here but the static stuff is not.

drphilmarshall:
Absolutely. I originally wrote "pieces from DC1 and Twinkles" but I couldn't find a specific DC1 cookbook recipe (markdown file). I thought the DC1 DM pipeline was derived from the Twinkles static sky recipe (that makes coadds, and objects=coaddsources), so figured that https://github.com/LSSTDESC/Twinkles/blob/master/doc/Cookbook/DM_Level2_Recipe.md would be a good place to start. The additional (or different) steps followed in the DC1 processing will then need to be captured, though - who would be best placed to add those parts?

katrinheitmann:
Just a reminder to deliver the first DC2 cookbook version.

heather999:
This issue was moved to LSSTDESC/ImageProcessingPipelines#15

PR #79, opened 2018-01-17
DC2 Cookbook, initial version

drphilmarshall:
@SimonKrughoff in response to #73 I have mashed together the three Twinkles Cookbook recipes, added a short overview section with a figure from LDM-151, and then flagged a few places where ACTION is needed. I think what's needed next is a careful reading from you to a) identify more to-do items, and b) check that we are describing two pipelines that really do emulate the expected DRP pipelines. Over to you, Simon! :-)

https://github.com/LSSTDESC/DC2_Repo/blob/issue/73/cookbook/Notebooks/DC2Cookbook.ipynb

> Sadly the navigation links don't work in GitHub's rendering - and its not picking up the images either... Best to check out the `issue/73/cookbook` branch and view the notebook locally, I think.

@cwwalter @jchiang87 Do you want to take a look at this now, to see what DC1 did differently, and/or suggest modifications via this thread? Thanks!

@nicolaschotard @boutigny While this first draft is still pretty rough around the edges, if you were able to give it a quick read and let us know if there are any _high-level_ changes you would recommend (or significant things missing), that could be most helpful. Simon and I will be working on cleaning up the details and adapting them to DC2, so we may quickly get to a point where we miss the "big picture"... Thanks!


jchiang87:
I've only scanned the content so far, but I note immediately that the `astrometry.net` stuff has been superseded. I think the correct procedure is captured [here](https://github.com/LSSTDESC/ImageProcessingPipelines/blob/master/bin/generate_dc_reference_catalog.py) and [here](https://github.com/LSSTDESC/ImageProcessingPipelines/blob/master/python/desc/parsl/pipeline_components/pipeline_components.py#L120)

jchiang87:
I think it would be useful to include a link to the doxygen documentation that describes the various tasks used in the multiband coadd processing steps:

http://doxygen.lsst.codes/stack/doxygen/x_masterDoxyDoc/pipe_tasks_multi_band.html

cwwalter:
Yes, as Jim says, in DC1 we originally were starting with things like:

https://github.com/LSSTDESC/Twinkles/blob/Issue%23435/updatecookbook/doc/Cookbook/DiaObjectGen.md#common-operations-for-both-l2-and-l1-type-processing

But, we were having astrometry problems and at that point the solution was that Simon help transition everything over to the new system.  I can't remember if this was documented in the same way. I remember Simon pointing me at some DM documentation but don't seem to have it bookmarked. On the DC1 side, in the end, this was implemented by Simon talking to Tony.

jchiang87:
I'd recommend removing the section on building the stack via lsstsw.  In addition to being hopelessly out of date, getting the Stack installed should be a separate recipe.

jchiang87:
The implementation of the reference catalog generation I pointed to is based on what Tony did for DC1 and on the QA work that Lauren and Heather worked on.

jchiang87:
We need steps describing the ISR if we are really intent on using the amplifier images in the pipeline.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -0,0 +1,559 @@
...
+    "\n",
+    "At the time of writing, multi-epoch object characterization is not yet available in the DM stack. As a result, the `Objects` are the same as the \"Preliminary Objects\", and are made by detecting sources in the `Coadd` images.\n",
+    "\n",
+    "Note that there will be difference image analysis (DIA) carried out in both the Alert Production and Data Release Processing pipelines. The main difference is that the APP will need to build `DIAObjects` on the fly, while the DRP will have all data available to it at once, after the fact. The DRP can make the best possible `TemplateCoadd` reference image, and then generate new `DIASources` and aggregate them into new, \"reprocessed\" `DIAObjects` ready for a single pass of forced photometry at their locations of all these reprocessed `DIAObjects`, which will result in light curves extending over the duration of the whole survey to that point.  As in Twinkles, in DC2 our goal for the dynamic sky analysis is to emulate the production of cosmology-ready LSST supernova and lensed quasar light curves, which is why we retain the focus on the _data release reprocessing_ of the visit images."
```
SimonKrughoff: I think we should probably mention forced photometry on per epoch images seeded by the objects detected from the coadd.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -0,0 +1,559 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<a id='coadds'></a>\n",
+    "## Image Coaddition and `Object` Detection\n",
+    "\n",
+    "This recipe started life as the original Twinkles Cookbook recipe, [_\"Recipe: Emulating the DM Level 2 Pipeline\"_](https://github.com/LSSTDESC/Twinkles/blob/master/doc/Cookbook/DM_Level2_Recipe.md). This was intended to show how to process simulated image data as if we were running the static sky part of the LSST DM DRP. The primary products are catalogs of `Objects.`"
```
SimonKrughoff: I think we should state that the main products are Objects and Forced photometry.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -0,0 +1,559 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Build the indexes for astrometric and photometric calibration\n",
```
SimonKrughoff: This should be replaced with Jim's pointers to the new style indexing.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -0,0 +1,559 @@
...
+   "metadata": {},
+   "source": [
+    "### Set up the data to run DM processing\n",
+    "\n",
+    "First you'll need to build the stack using tickets/DM-4302 of obs_lsstSim.  In order to patch a branch version onto a pre-existing stack you can do something like the following:\n",
+    "\n",
+    "1. Build a master stack.  I suggest using [lsstsw](https://confluence.lsstcorp.org/display/LDMDG/The+LSST+Software+Build+Tool).\n",
+    "2. Set up the stack: e.g. `$> setup obs_lsstSim -t bNNNN`\n",
+    "3. Clone the package you want to patch on top of your stack `$> clone git@github.com:lsst/obs_lsstSim.git; cd obs_lsstSim`\n",
+    "4. Get the branch: `$> checkout tickets/DM-4302`\n",
```
SimonKrughoff: I wonder if we should make a DC2 branch of `obs_lsstSIm`.  We didn't do this for Twinkles or DC1 and it resulted in problems.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -0,0 +1,559 @@
...
+   "metadata": {},
+   "source": [
+    "### Process the image data using the DM stack\n",
+    "\n",
+    "Start here if you just want to exercise the DM stack.  If you didn't follow the steps above, first get the data and astrometry.net index files from\n",
+    "[here](https://lsst-web.ncsa.illinois.edu/~krughoff/data/gri_data.tar.gz).  Then untar the tarball in a working directory.\n",
+    "\n",
+    "After you have the data, you can start following the steps below to get, for example, forced photometry in three bands. First, set up the reference catalog for photometric and astrometric calibration:\n",
+    "```bash\n",
+    "$> setup -m none -r and_files astrometry_net_data\n",
```
SimonKrughoff: Again, we can update for the newer system.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -0,0 +1,559 @@
...
+    "```"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Create calibrated images from the input eimages.  This will write to a repository called output_data.  The --id argument defines the data to operate on.  In this case it means process all data (in this example the g, r, and i bands) with visit numbers between 840 and 879.  Missing data will be skipped.\n",
+    "```bash\n",
+    "$> processEimage.py input_data/ --id visit=840..879 --output output_data\n",
```
SimonKrughoff: One problem we ran into last time was that I specified data ids in the cookbook that weren't the same as the ones for production.  This was because I just wanted to reduce a manageable set and I copied the command lines into the cookbook verbatim.  It would be good to try to figure out how id can be specified for both a quick run and for a large scale production.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -0,0 +1,559 @@
...
+    "```bash\n",
+    "$> forcedPhotCcd.py output_data/ --id tract=0 visit=840..879 sensor=1,1 raft=2,2 --config measurement.doApplyApCorr=yes --output output_data\n",
+    "```"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This final step is not really necessary: it results in a `ForcedSource` table whose utility is questionable. We expect the light curves of supernovae to come from the forced photometry of the `DIASources` (see below). However, the forced photometry of the static sky `Objects` may provide some useful comparisons, so we include it.\n",
```
SimonKrughoff: I think we can soften the words here a little bit.  The ForcedSources will be useful, if just for gauging the performance.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -0,0 +1,559 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Aggregate quantities for `DIAObjects`\n",
+    "We will take the associated catalog from `MultiMatch` and compute aggregate quantities for the columns that impact the forced photometry: i.e. positions, flags, and the total number of `DIASources` associated with the `DIAObject`.  The aggregate quantites will be persisted in a new dataset `reproDIAObjects`.\n",
+    "\n",
+    "**ACTION: Update the above once new association code is included. Maybe can be removed?**"
```
SimonKrughoff: I don't think the new code will do the property aggregation.

Notebooks/DC2Cookbook.ipynb
jchiang87: I'd prefer that DESC have more direct control over the changes that would need to be made in obs_lsstSim.  I don't think that working off of branches of a repo in the lsst org gives us the control we need.  I would propose instead forking obs_lsstSim into LSSTDESC.  That will give us the freedom to control the development workflow, i.e., develop along whatever branches we want, apply release tags to master whenever we want, and thereby maintain better provenance of the code we use in our production runs.  We can always PR from our fork back to the lsst version.

drphilmarshall:
Thanks a lot, @jchiang87 ! I've issued the rest of your feedback, as well as the missing pieces I could see myself, and turned #73 into a (ZenHub) epic, to keep the work together.

nicolaschotard:
@drphilmarshall I've had a quick look at the cookbook and was thinking that the sky map that we will make should be absolute (using `makeSkyMap.py` and the appropriate configuration, e.g. for a ring sky map as for HSC and now CFHT reprocessing) instead of relative to the field (with `makeDiscreteSkyMap.py`). It would ensure that the sky map won't depend on the list of input visits, and thus on the time and place the reprocessing happens.

boutigny:
There is nothing in the cookbook related to `ingestImages`. I also see that there is an `ingestSimImages.py` in `lsst_sims` which apparently contains some fixes for keywords which are missing in galsim generated images.
I think that it is a little bit dangerous to have customized versions of standard scripts. It would be better to fix the fits headers of simulated images.

drphilmarshall:
Thanks @nicolaschotard , @boutigny ! I just issued these suggestions (#121, #122). 

@SimonKrughoff @nicolaschotard @boutigny @DominiqueFouchez @jchiang87 We have a fair number of bite-sized DC2 Cookbook issues that need fixing, via collaborative editing of the [Cookbook Notebook](https://github.com/LSSTDESC/DC2_Repo/blob/issue/73/cookbook/Notebooks/DC2Cookbook.ipynb). I propose we all look through the list (either via https://github.com/lsstdesc/dc2_repo#boards?epics=103439116_73&repos=103439116 if you have ZenHub installed, or via the [Cookbook label](https://github.com/LSSTDESC/DC2_Repo/issues?q=is%3Aissue+is%3Aopen+label%3ACookbook) and self-assign whatever we can volunteer to take on. The goal is to produce a clear description of a complete DRP pipeline that is consistent with Project specifications and documentation. Whatever edits we need to make to get that, we need to do them!  I'll post this to Slack as well. Thanks for your help with this!

Notebooks/DC2Cookbook.ipynb
```diff
@@ -393,19 +393,35 @@
...
+    "\n",
+    "One possibility could be to use the [Rings sky map](https://github.com/lsst/skymap/blob/master/python/lsst/skymap/ringsSkyMap.py#L42), as done for the HSC data reprocessing with the LSST stack (and currently for CFHT data reprocessing as well). In order to make this sky map, some modifications of the `lsstSimMapper.py` [script](https://github.com/lsst/obs_lsstSim/blob/master/python/lsst/obs/lsstSim/lsstSimMapper.py) must be done to match what has been done in `hasMapper.py` ([here](https://github.com/lsst/skymap/blob/master/python/lsst/skymap/ringsSkyMap.py#L42) and [here](https://github.com/lsst/obs_subaru/blob/master/python/lsst/obs/hsc/hscMapper.py#L286)). A configuration file (`makeSkyMapConfig.py` below) is then needed to run `makeSkyMap.py`, containing information on how to parametrize the rings map. An example of such a configuration can be taken from the HSC reprocessing [repository](https://github.com/LSSTDESC/ReprocessingTaskForce/blob/master/config/w_2017_49/cfht/makeSkyMapConfig.py), which contains:\n",
```
SimonKrughoff: We will need to make this a concrete suggestion rather than a statement of what is possible.  Can we make such a suggestion, with associate config, now, or do we need to have more info?

nicolaschotard: I was waiting for reaction from different people and left the choice of the sky map opened on purpose. But I think that using a ring sky map is reasonable, with a configuration close to what was use for HSC.

SimonKrughoff: Agreed.  I'm fine closing this with this wording, but we will need to finalize it at some point.

wmwv:
@drphilmarshall @SimonKrughoff Is this branch still worth merging in?

jchiang87:
I don't think it's relevant anymore.  It's been superseded by the wiki over in https://github.com/LSSTDESC/ImageProcessingPipelines .
I would close without merging.

wmwv:
Closing without merging.

PR #85, opened 2018-01-18
Documents README with links to DC2 Plan and Run 1.1 Specs

drphilmarshall:
Looking to make it easier to find our documentation - I'll work up the main README as well, and then assign Katrin as reviewer.


drphilmarshall:
OK @katrinheitmann I think this is ready for review! In the meantime, I'll go check the paper tracking table on confluence to make sure the two DC2 Notes appear there. Thanks!

PR #123, opened 2018-01-24
descDithers +  protoDC2 visit list generation code and output files.

humnaawan:


rbiswas4:
@jchiang87  What is the workflow for PRs to DC2_repo. Humna has this on her fork, it should be fine to merge directly from there, right? She also has a few data products which are small ~ 516KB. These are intermediate products that are important for DC2 simulation inputs (I believe you only want the inputs in  Should we maintain this in the github repo or keep them on NERSC, and have them documented in the tracking file?

rbiswas4:
@humnaawan Aside from intermediate products, it would be good to keep track of packages that we are using to create these products (in the same spirit as @jchiang87 is keeping packages that are actually being used for DC2 simulations and analysis in config. I am happy to put these in his config if he wants, but I don't think he would, but we should maintain them for survey design purposes). You used your [DC1Regions](http://github.com/humnaawan/DC1-Regions) package to generate the list of pixels. Could you please create a release tag (this involves going to the [release button](https://github.com/humnaawan/DC1-Regions/releases) and following the instructions and should be super quick). I did the same for [opsimsummary](https://github.com/rbiswas4/OpSimSummary) just now, for the same reason. 

Also I want to add a couple of script as part of the PR to bridge the gap on a few steps. Let us see what Jim says on how to merge, and I would like to figure out how best to do this.

jchiang87:
I think we should only have data files in the repo that will actually be used at NERSC for the phosim runs.  As we discussed in [today's version of the Run 1.1 checklist](https://docs.google.com/document/d/10PBxkR0ekDrTmeG_N1ot5B0q2ML0OUd6Xc31PmBptPo/edit?usp=sharing), we need to have the WFD and uDDF visit lists broken down in to separate lists for the six bands, following the format used for Run1.0b.   Any other files that can be regenerated from the code (e.g., via the notebook) I don't think belongs in this repo.

humnaawan:
@rbiswas4 so I've removed the dependency on DC1 code; there were only two functions and I was modifying their outputs anyway. Also, I have added to the ipython notebook additional info you requested.

Since there isnt much point in keeping all interm code/products for the visit list generation, I'll move all my code (with its commit history) to a separate repo; similar to the DC1 repo from where we only access the data files. I can tag and release that, which would allow tracking the specific output we end up using. Does that sound ok?

rbiswas4:
@humnaawan That sounds like a good idea. Can you point me to the repository with the extra stuff, I have some things like that as well, and it would be good to put all of these products together so that we don't forget.

jchiang87:
It looks like there are now zero net changes in this PR.  In that case, I will close it now.  Please re-open (or file a new issue) if more work along these lines needs to be done.

PR #124, opened 2018-01-25
First attempt to describe how to make the sky map

nicolaschotard:


Notebooks/DC2Cookbook.ipynb
```diff
@@ -393,19 +393,35 @@
...
+    "\n",
+    "```\n",
+    "config.skyMap.name = \"rings\"\n",
+    "config.skyMap[\"rings\"].numRings = 120\n",
+    "config.skyMap[\"rings\"].projection = \"TAN\"\n",
+    "config.skyMap[\"rings\"].tractOverlap = 1.0/60 # Overlap between tracts (degrees)\n",
+    "config.skyMap[\"rings\"].pixelScale = 0.185\n",
+    "```\n",
```
drphilmarshall: @jchiang87 : in his new `makeSkyMap` cookbook recipe section, @nicolaschotard is suggesting we adapt an LSST script, as HSC have done, and provide a configuration, analogous to the CFHTLS repro one. Can you suggest a sentence of the form "this modified script and configuration should live in the DESC repo XXX" - I'm not sure what we should be forking. Nicolas, feel free to suggest a sensible scheme, too.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -393,19 +393,35 @@
...
-    "Input to the sky map maker should reflect all of the data to be coadded in the future.  This is typically done by specifying the `--id` flag with no argument.  The sky map should be generated once and used by all the bands.\n",
+    "After having adapted `lsstSimMapper.py` and getting the configuration file ready, the command to run will be:\n",
+    "    \n",
     "```bash\n",
-    "$> makeDiscreteSkyMap.py $CALEXPDIR --id --output $COADDDIR\n",
-    "```"
+    "$> makeSkyMap.py INPUTDIR --output OUTPUTDIR --configfile makeSkyMapConfig.py\n",
+    "```\n",
+    "\n",
+    "The list of tracts/patches in which there is actually data (out of the 18937 tracts in that case) can be determined using a [hand-made script](https://github.com/LSSTDESC/ReprocessingTaskForce/blob/master/scripts/reportPatchesWithImages.py) developped in the context of CFHT data reprocessing (that might need some improvement)."
```
drphilmarshall: Maybe best to bring a copy of this home-grown `reportPatchesWithImages.py` script into the ImageProcessingPipelines repo? Or is `import desc.reprocessingtaskforce` a viable command? cc @jchiang87 again.

Notebooks/DC2Cookbook.ipynb
```diff
@@ -393,19 +393,35 @@
...
+    "\n",
+    "One possibility could be to use the [Rings sky map](https://github.com/lsst/skymap/blob/master/python/lsst/skymap/ringsSkyMap.py#L42), as done for the HSC data reprocessing with the LSST stack (and currently for CFHT data reprocessing as well). In order to make this sky map, some modifications of the `lsstSimMapper.py` [script](https://github.com/lsst/obs_lsstSim/blob/master/python/lsst/obs/lsstSim/lsstSimMapper.py) must be done to match what has been done in `hasMapper.py` ([here](https://github.com/lsst/skymap/blob/master/python/lsst/skymap/ringsSkyMap.py#L42) and [here](https://github.com/lsst/obs_subaru/blob/master/python/lsst/obs/hsc/hscMapper.py#L286)). A configuration file (`makeSkyMapConfig.py` below) is then needed to run `makeSkyMap.py`, containing information on how to parametrize the rings map. An example of such a configuration can be taken from the HSC reprocessing [repository](https://github.com/LSSTDESC/ReprocessingTaskForce/blob/master/config/w_2017_49/cfht/makeSkyMapConfig.py), which contains:\n",
+    "\n",
```
drphilmarshall: I think here we need to cite the CFHTLS example but use the LSST values, eg `pixelScale = 0.20`. Who is best placed to advise on the values of `tractOverlap` and `numRings`? I think we can merge using your values as defaults, but we should target a quick question at the DM team to check unless you think the ones below _are_ what LSST would use.

nicolaschotard: I gave this configuration file as an example, and we should definitely ask someone from DM to have a look at it and validate our choice of sky map. Maybe @TallJimbo or @PaulPrice might have a recommendation about that.

Notebooks/DC2Cookbook.ipynb
nicolaschotard: The configuration will follow the choice of sky map we will use. What will have to be updated accordingly to this choice is the `lsstSimMapper.py` script that leaves in the `obs_lsstSim` github repository ([here](https://github.com/lsst/obs_lsstSim/blob/master/python/lsst/obs/lsstSim/lsstSimMapper.py)). All changes should actually be done in `obs_lsstSim`, and the configuration file should also be put in there as a default configuration for the sky map. But again, DM people might have a different view on that.

Notebooks/DC2Cookbook.ipynb
nicolaschotard: `import desc.reprocessingtaskforce` won't work, but we can definitely put this script in the ImageProcessingPipelines.

Notebooks/DC2Cookbook.ipynb
nicolaschotard: After a short discussion about that on the slack `dm` channel with @TallJimbo, the only modification that must be done is for the `pixelScale`, as @drphilmarshall mentioned above. The number or rings could be decrease a little to make the tracts bigger and better match LSST, but that less important.

Notebooks/DC2Cookbook.ipynb
nicolaschotard: I opened an (Jira ticket)[https://jira.lsstcorp.org/browse/DM-13377] about that ().

Notebooks/DC2Cookbook.ipynb
PaulPrice: You _might_ also want to adjust the `tractOverlap`. This sets the maximum size you can reliably extract from a single image. If you want to be able to extract NGC galaxies without having to warp tracts, you may want to make it bigger. If you only care about point sources and want to save disk space, you could make it smaller.

nicolaschotard:
A DM [RFC](https://jira.lsstcorp.org/browse/RFC-438) has been opened by @TallJimbo on that topic.

PR #129, opened 2018-02-02
properly indicate involved working groups for a subset of projects

rmandelb:
This very small PR involves mild changes in section 3 of the DC2 planning document to properly indicate that some of the planned DC2 analyses listed as science drivers will have heavy involvement by more than one working group.  The relevant collaborations are (WL, TJP) and (LSS, TJP) and (WL, LSS, TJP) as indicated in the document.  This addresses the issue that TJP did not appear to be represented previously (whereas really it was implicitly there in some of the projects).

Flagging this to TJP conveners, @jablazek @elikrause @philbull


jablazek:
@rmandelb : Looks like a good way to structure these XWG tasks. Should IA be specifically mentioned in this section (it is one of the two main astrophysical systematics TJP is thinking about for DC2)? Or is it sufficient to discuss that in more detail later in the document?

rmandelb:
@jablazek - The purpose of the section is to briefly give a sense of the science that will come out of DC2, but I think going into more detail about the astrophysical systematics would lead to inconsistencies among the projects that are mentioned.  We are not explicitly going into further details of these projects later in the document either; however, if there are points at which IA is mentioned as a driver for some aspect of DC2 (e.g. because it leads to a need for different data products) I think it can simply be explained at that point.

Issue #108, opened 2018-01-20
Visit IDs that overlap with WFD and uDDF region


rbiswas4:
We would like to have  two lists of visits from the Run 1.1 `minion_1016_desc_dithered_vn.db` #98 by  Monday Jan 22  6pm. 
- [ ] `AllVisitsWFD_vn.csv` which is a csv file which has the set of visits (`obsHIstID`) from the above database that overlap with the main survey region for Run 1.1. Requires #98 
- [ ] `AllVisitsuDDF_vn.csv` which is a csv file which has the set of visits (`obsHIstID`)from the above database that overlap with the uDDF region for Run 1.1.  Requires #96 #98 

If this is trivial because the dithers are small enough (eg. in uDDF and the field is in the center, that is fine, let it be a trivially composed list)

Why do we want this ? The visit lists needed for Run 1.1 must be subsets of these (default would be the first set of observations from this subset) . 

@humnaawan probably has code ready to trivially do this. (please say if there is a problem for example either required inputs missing or something else.)

humnaawan:
@rbiswas4 I didnt follow the naming scheme above since it ignores some of the things that were important for me to track their production. So `AllVisitsWFD_vn.csv` would be `DC2VisitList_minion1016_v3_WFDvisits_nside512_WFDvisitsInWFD.csv`, and `AllVisitsuDDF_vn.csv` would be `DC2VisitList_minion1016_v3_WFDvisits_nside512_WFDvisitsInuDD.csv`. 

Hope this wont be a big problem.

humnaawan:
Also, here are GIFs that illustrate the three files that are in the data folder:

![hids_0-100index_udd_test_nside512](https://user-images.githubusercontent.com/12702187/35362959-a5ab9292-0135-11e8-80b5-e74ef4a1e319.gif)

^(=> The uDDF dithers working.)

Theres are first 100 WFD visits:
![hids_0-100index_wfd_test_nside512](https://user-images.githubusercontent.com/12702187/35362960-a5b3c322-0135-11e8-93d8-2ed69774a3e9.gif)

These are first 100 WFD visits that are in the uDD footprint:

![hids_0-100index_wfdinudd_test_nside512](https://user-images.githubusercontent.com/12702187/35362961-a5bd7fb6-0135-11e8-95c6-5ceca0d2e78b.gif)

drphilmarshall:
[...]

PR #132, opened 2018-02-10
Issue131/run1.1 visit list

### LINKED ISSUE #131: Run 1.1 Visit List
@rbiswas4 looking over the [notebook](https://github.com/LSSTDESC/DC2_Repo/blob/master/Notebooks/SurveyDesign/get_visit_lists.ipynb) that produces the final list for the visits, it appears that we are not using the correct WFD visits: `DC2VisitList_minion1016_v3_WFDvisits_nside512_WFDvisitsInuDD.csv` has WFD visits that overlap with the uDD region .. not what we want for Run 1.1 which needs `DC2VisitList_minion1016_v3_WFDvisits_nside512_WFDvisitsInWFD.csv`.

I'm sorry I didnt catch this earlier but I misread Issue #98 ("Visit IDs that overlap with WFD and uDDF region") to mean visits that overlap both WFD and uDDF and thought the `WFDvisitsInuDD.csv` file use was intentional. We dont really even need that file, do we?

rbiswas4:
However, we should write a test (but in a separate issue).

fixes #131


katrinheitmann:
So we are waiting for @humnaawan to check this? @jchiang87 You are aware of this, right? I just wanted to make sure this change was not missed. Sorry if all of this is under control already ;-)

jchiang87:
@rbiswas4 Sure, I saw this.  I think for Run1.1p, getting this fix in isn't crucial. At least that is what you and I concluded during a brief discussion last week.  If people think this is essential for Run1.1p, then we definitely need @humnaawan or someone else to review this asap.

rbiswas4:
@jchiang87 Yes, if we had not found this problem I think we would have still learned from Run 1.1p, but having found the problem, I would like to make sure this is Run 1.1p and we get what we designed it for.

Notebooks/SurveyDesign/get_visit_lists.ipynb
```diff
@@ -56,42 +52,36 @@
...
   {
    "cell_type": "code",
    "execution_count": 5,
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "opsimdb = '/Users/rbiswas/doc/projects/DESC/Notes/DC2_visits/minion_1016_desc_dithered_v4.db'"
```
humnaawan: So whats this v4? It is not mentioned in #108, and my outputs were produced using the v3 database.

Notebooks/SurveyDesign/get_visit_lists.ipynb
```diff
@@ -151,9 +135,7 @@
   {
    "cell_type": "code",
    "execution_count": 9,
-   "metadata": {
-    "collapsed": false
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "def write_visitlists_uddf(csvfname, opsimout, NumVisits=NumVistitsuDDF*19 + 170, rootfname='../../data/Run1.1/protoDC2_visits_uDDF',\n",
```
humnaawan: `rootfname` will need to be changed now that the relevant csv files are sitting in a separate [repo](https://github.com/LSSTDESC/DC2_visitList).

Notebooks/SurveyDesign/get_visit_lists.ipynb
humnaawan: @rbiswas4 can you comment on this? This v4 is the only thing that I am not sure about in your notebook.

rbiswas4: About v4: If you remember, there was a problem with v3, because in the v3 I put did not have the usual indices (not the really big deal), but somehow it got corrupted/copied over or something, and I was not able to get the same results you were getting from the v3 on nersc. I then put v4 which was v3 (with the indices in place) and put them on nersc and checked that this and your version of v3 matched. v4 is what is on the Run 1.1 specifications  document.

humnaawan: Oh ok. I guess we might need to think about minimizing the versioning but as long as things are documented, we'd hopefully not run into too much trouble.

PR #146, opened 2018-03-26
Run1.2p configuration

jchiang87:
@katrinheitmann @TomGlanzman Here are my guesses at our planned configuration for Run1.2p.  Please have a look, make corrections, and loop in anyone else who should help specify the configurations.

### DIFF
--- a/config/prod_software_config.ini
+++ b/config/prod_software_config.ini
@@ -1,5 +1,5 @@
 [config]
-DC2_Repo = v0.2.0
+DC2_Repo = 
 [simulation]
 descqa = v2.0.0-0.6.0
 gcr-catalogs = v0.6.14
@@ -7,7 +7,7 @@ Twinkles = DC2-Run1.1-v4
 drphilmarshall/OM10 = 0.8
 drphilmarshall/LensPop = 0.8
 sims_GCRCatSimInterface = v0.1.1
-PhoSim = v3.7.8
+PhoSim = v3.7.9
 GalSim-developers/GalSim = v1.5.1
 imSim =
 DC2-phoSim-3 = v1.0.0
--- a/config/prod_software_config.ini
+++ b/config/prod_software_config.ini
@@ -15,8 +15,8 @@ DC2-phoSim-3 = v1.0.0
 ImageProcessingPipelines =
 obs_lsstSim =
 [stack_versions]
-lsst_distrib = w_2018_03
-lsst_sims = sims_2_5_0
+lsst_distrib = w_2018_09
+lsst_sims = sims_2_6_0
 [nersc_files]
 opsim_db = /global/projecta/projectdirs/lsst/groups/SSim/DC2/minion_1016_desc_dithered_v4.db
 agn_db = /global/projecta/projectdirs/lsst/groups/SSim/DC2/agn_db_mbh_7.0_m_i_30.0_gcr_protodc2_v3.db
--- a/config/prod_software_config.ini
+++ b/config/prod_software_config.ini
@@ -2,7 +2,7 @@
 DC2_Repo = 
 [simulation]
 descqa = v2.0.0-0.6.0
-gcr-catalogs = v0.7.0
+gcr-catalogs = v0.7.1
 Twinkles = DC2-Run1.1-v4
 drphilmarshall/OM10 = 0.8
 drphilmarshall/LensPop = 0.8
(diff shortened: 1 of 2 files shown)


katrinheitmann:
@rmandelb @rmjarvis and anybody else: the difference between 33 and 34 is clouds -- I personally prefer tests to be easier so would keep them off and follow Jim's suggestion above (setting for 33) but would like to hear what others think. Are there any more quick tests we can do to decide? Thanks!

katrinheitmann:
@yymao @evevkovacs @danielsf With Run 1.2p it would be good to switch now to protoDC2_v3. Can you please help/make sure that we have all the correct version numbers for the GCR, catalog, and also DESCQA in the configuration set-up? I know that Yao helped with that for Run 1.1p. Thanks a lot!

yymao:
@katrinheitmann @jchiang87 I've updated the gcr-catalog version in `config/prod_software_config.ini`.   Once the production environment is updated, it'll automatically switch to v3.

katrinheitmann:
Thanks a bunch!!

jbkalmbach:
@jchiang87 Do you know where you will be rotating the center of the v3 catalogs so that we can make sure the sprinkler cache of galaxy_ids is updated?

jchiang87:
@jbkalmbach The most recent info that I know of on that question is the discussion in the #desc-dc2 channel on Friday that you prompted.  Phil's response in that thread seemed to refer to cosmoDC2, not protoDC2_v3, so assuming protoDC2_v3 and Run1.2p will cover the same 5x5 degree region, I would assume that the field and rotation will be the same as for Run1.1p.   If someone knows differently, please correct this information!

katrinheitmann:
My understanding as well (field and rotation will not change). Just one 
thing: the galaxy positions in v2 and v3 will not be identical (the 
clustering properties were improved). So if you need to know the 
positions of the galaxies and you get them from the instance catalog, it 
will need to be regenerated. This remark might be completely unnecessary 
and you are starting from something else, but just wanted to make sure 
you are aware of this.

jbkalmbach:
@jchiang87 @katrinheitmann Thanks. Yes, I will have to start from the new v3 catalog to get everything I need (ra, dec, galaxy_id, ...) for the galaxies in the uDDF field so everything should be consistent with the new version when I update.

cwwalter:
I agree keeping clouds off for 1.2 is safest as it creates effects similar to those everyone is focused on testing.

jchiang87:
Everything looks good to me.

sethdigel:
Regarding

```
# Use defaults for back* commands
#backalpha 0.1
#backbeta 4.0
#backgamma 1000.0
#backdelta 1.0
```

This looks fine, in the sense that the entries are all commented out so the parameters end up with their default values.  What you've listed as the commented-out values are the defaults except for backgamma (which defaults to 2).

cwwalter:
This all looks OK to me.

There is one funny bug I once came across in PhoSim. I can't remember the command now but even if the line was commented out the value was getting picked up. I remembering telling John about it.  So, given how much CPU you are about to burn; if you don't remove these I would double check by looking at the header of a test file and make sure it does in fact I have what you want.

jchiang87:
@cwwalter It sounds like if we just removed those commented out commands, we'd be safe, right?

cwwalter:
> @cwwalter It sounds like if we just removed those commented out commands, we'd be safe, right?

Right.  The problem was the parser was picking up the values in the line even it it was commented out.  If you just remove them it would be fine.

jchiang87:
Hi All,

It looks like we are nearly ready, software-wise, for Run1.2p.  As of this morning, we have a sims version (v2.7.0) that we will use, and that is being built now at NERSC; and all of the other packages that need updates, except one, have been tagged and their versions entered here.

The only remaining package is `sims_GCRCatSimInterface`, which has an open pull-request waiting for review, https://github.com/LSSTDESC/sims_GCRCatSimInterface/pull/41.   @danielsf @yymao:  Can you guys work on getting that merged? (FYI:  I tried the new code and it produced identical catalogs as the current code).  Once that is merged, I'll tag that package and enter that version here.

After that is done, I'll ask Tom and Katrin to do the review for this PR.

If there are any other outstanding issues that anyone can think of, please chime in.

jchiang87:
@katrinheitmann @TomGlanzman I think everything is good to go.  Can you both give this set of tags a quick look and if all looks good, mark the PR approved?  Thanks.  Everything is installed at the prod area except this repo, which I will install once it's tagged.

PR #153, opened 2018-04-30
Issues/151 - Add script to generate merged static-sky coadd catalogs as HDF5 files.

wmwv:
Adds `merged_tract_cat.py` and `example_merge_tract_cat.py` to `script` directory.

Creates one HDF5 file per tract with all ~2400 columns from coadd merged reference catalogs + 6 filters. 

Specific questions to reviewers:
* Look in `/global/projecta/projectdirs/lsst/global/in2p3/Run1.1-test2/summary` on NERSC for summaries generated from the Run 1.1p process done at IN2P3 and transferred to NERSC.  Load some catalogs and try things out.
* Look in `examples_merge_tract_cat.example_load_and_plot` for an example use and plotting.

1. [ ] Should the HDF5 files have a separate HDF Groups for each patch or just one Group for the entire tract?  I'm somewhat new to HDF5 -- is there a third, better way?
2. [ ] If separate groups per patch is there an easy way for the user who just wants everything to grab that without having to know the names of group identifier?
3. [ ] If separate groups per patch, should they be named `coadd_4849_1,1` or `coadd_4849_11`.  The latter would allow the identifier to be used as an attribute because it's a valid Python identifier.
4. [ ] Should there be a light-weight HDF5 file with only the most useful columns?  If so, should this be separated by patch as well for consistency (if we decide to go that way), or should it be merged across patches for convenience?
5. [ ] `example_merge_tract_cat.py` presents a list of examples as functions and then has them turned on/off by booleans in the main run.  Is there a better way to present these examples?

PR #157, opened 2018-05-01
GCR reader for DC2 static coadd catalogs

yymao:
This PR adds a GCR reader for the DC2 static coadd catalogs generated by @wmwv in #153 and also [an example notebook](https://github.com/LSSTDESC/DC2_Repo/blob/u/yymao/GCRforDC2Coadd/Notebooks/DC2%20Static%20Coadd%20Catalog%20reader.ipynb) of how to use it.


wmwv:
The LSST Project refers to its 'y' filter as lower-case, and I presume this will be the case for distributed data products as well.

RobertLuptonTheGood:
HSC also calls it y.   Y is usually reserved for the HgCdTe c. 1 micron band.

wmwv:
Expand patches back out as strings.

In the LSST current skymap scheme patches are strings.  E.g., '0,4', 1,2' or '2,3'.  It is often convenient to represent them as '04', '12', or '23', but those shouldn't be translated back into ints of 4, 12, 23.  Such ints are of course unique, but it's less obvious how to get all of the patches in a given row or column and makes the neighbor relationship less clear.

yymao:
thanks for the feedback!

scripts/reader/dc2_static_coadd_reader.py
```diff
@@ -10,36 +10,45 @@
 
 class DC2StaticCoaddCatalog(BaseGenericCatalog):
 
-    _native_filter_quantities = {'tract', 'patch'}
+    _native_filter_quantities = {'tract', 'patch_ra', 'patch_dec'}
```
wmwv: I don't like patch_ra and patch_dec, although I'm fine with dividing them up.  There's nothing that says that patches have to be oriented along RA or Dec.

wmwv: Do you need an int lookup for something?  Is there a reason you can't just have
```
_native_filter_quantities = {'tract', 'patch'}
```
but have patch be a str?

yymao: @wmwv it surely can. I've updated per your suggestions. The only downside is that people cannot do `(lambda p: p < 2, 'patch_ra')` as a native filter, but I agree with you that this is probably not a common use case.

wmwv:
@yymao Could you merge this in in advance of the DESC Seminar tomorrow (May 30)?

yymao:
Sure. If there's no objection I'll merge this tonight. 

(cc @jchiang87 @katrinheitmann @cwwalter)

PR #172, opened 2018-06-26
Fix isPrimary bug for empty patches

wmwv:
Patches with no valid entries marked 'isPrimary' in ref_table
were returning _all_ ref_table entries.  This fixes that by
  * Explicitly restricting ref_table to isPrimary
  * Skipping patches that are empty lists.

### DIFF
--- a/scripts/merge_tract_cat.py
+++ b/scripts/merge_tract_cat.py
@@ -1,7 +1,7 @@
 import re
 import sys
 
-from astropy.table import join, vstack
+from astropy.table import join, vstack, Table
 from astropy.utils.metadata import MergeStrategy, enable_merge_strategies
 import numpy as np
 
@@ -88,12 +88,10 @@ def load_and_save_tract(repo, tract, filename, key_prefix='coadd', patches=None,
     for patch in patches:
         if verbose:
             print("Processing tract %d, patch %s" % (tract, patch))
-        try:
-            patch_merged_cat = load_patch(butler, tract, patch, **kwargs)
-        except NoResults as e:
-            print(e)
-            continue
+        patch_merged_cat = load_patch(butler, tract, patch, verbose=verbose, **kwargs)
         if len(patch_merged_cat) == 0:
+            if verbose:
+                print("  No good entries for tract %d, patch %s" % (tract, patch))
             continue
 
         key = '%s_%d_%s' % (key_prefix, tract, patch)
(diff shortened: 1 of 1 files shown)


scripts/merge_tract_cat.py
```diff
@@ -93,6 +93,8 @@ def load_and_save_tract(repo, tract, filename, key_prefix='coadd', patches=None,
         except NoResults as e:
             print(e)
             continue
+        if len(patch_merged_cat) == 0:
+            continue
```
yymao: Maybe add `print('empty patch')` so that it has a similar behavior like `print(e)` above?

scripts/merge_tract_cat.py
```diff
@@ -126,6 +128,8 @@ def load_tract(repo, tract, patches=None, **kwargs):
         except NoResults as e:
             print(e)
             continue
+        if len(this_patch_merged_cat) == 0:
+            continue
```
yymao: Maybe add `print('empty patch')` so that it has a similar behavior like `print(e)` above?

wmwv: 👍 
I got a bit more elaborate with standardizing the error messages in my final changes.  Thanks for the suggestions.
So now it outputs things like
```
Processing tract 4433, patch 7,4
  No locations for get: datasetType:deepCoadd_ref dataId:DataId(initialdata={'tract': 4433, 'patch': '7,4'}, tag=set())
  No good entries for tract 4433, patch 7,4
Processing tract 4433, patch 7,5
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 4433, 'patch': '7,5', 'filter': 'g'}, tag=set())
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 4433, 'patch': '7,5', 'filter': 'r'}, tag=set())
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 4433, 'patch': '7,5', 'filter': 'i'}, tag=set())
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 4433, 'patch': '7,5', 'filter': 'z'}, tag=set())
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 4433, 'patch': '7,5', 'filter': 'y'}, tag=set())
```

wmwv:
The tracts that I needed to regenerate were
4429 4433 4637 4640 4852 5066

scripts/merge_tract_cat.py
```diff
@@ -121,11 +121,12 @@ def load_tract(repo, tract, patches=None, **kwargs):
 
     merged_patch_cats = []
     for patch in patches:
-        try:
-            this_patch_merged_cat = load_patch(butler, tract, patch, **kwargs)
-        except NoResults as e:
-            print(e)
+        this_patch_merged_cat = load_patch(butler, tract, patch, **kwargs)
+        if this_patch_merged_cat == 0:
```
yymao: @wmwv do you need a `len` here? (Again, I didn't run the script, am just guessing the behavior of `load_patch`).

wmwv: Good catch.
But given that I didn't catch this before when running test examples right now, i realize that this test isn't even necessary.  It's fine to get empty `astropy.table.Table`s back and then vstack then to still an empty `Table`.  I've removed these lines.

PR #175, opened 2018-06-28
Issues/174

wmwv:
@EiffL Could you take a look at the demonstration Notebook for generating postage stamps.  
1) This is meant to be a demonstration, not a library, although I did try to write things out as civilized functions.  We need a better packaging structure in which to put library code.
2) I'd particularly appreciate guidance about how to clearly explain the purpose and potential utility of the Notebook.


EiffL:
I just had time to read through it (and I already learned a few cool tricks along the way :-) ), it's really cool, I'll give you a real review later today

RobertLuptonTheGood:
If you used afwDisplay to do the display you'd get automatic mask support, plus the ability to use ds9/ginga/firefly backends in addition to matplotlib.

wmwv:
@RobertLuptonTheGood That's a good point about the mask planes.  Those are an important thing to understand and for many of the use cases for postage-stamp visualization

I think I'll avoid explicit examples with ds9/ginga/firefly for now because I don't want to include the setup help that might be necessary for the average DESC member to get these going remotely from NERSC JupyterHub.  But writing it in afwDisplay would make that easier later on.

1. How do I save a PNG from an afwDisplay that has a matplotlib backend?
2. How do I get an inline display to show a list of plots instead of cycling through the same display matplotlib?  The `incrDefaultFrame` doesn't seem to work.  With the matplotlib afwDisplay backend, do I need an individual display for each plot?

PR #180, opened 2018-07-03
Add option to ingest HSC catalogs using the HDF5 script

EiffL:
@jchiang87 @wmwv I just added an `--hsc` option that changes the names of the filters to the HSC ones


scripts/merge_tract_cat.py
```diff
@@ -295,9 +295,15 @@ def prefix_columns(cat, filt, fields_to_skip=()):
...
 
     for tract in args.tract:
         filebase = 'merged_tract_%d' % tract
         filename = filebase+'.hdf5'
-        load_and_save_tract(args.repo, tract, filename, verbose=args.verbose)
+        if args.hsc:
+            load_and_save_tract(args.repo, tract, filename, verbose=args.verbose,
+                                filters=('HSC-U', 'HSC-G', 'HSC-R', 'HSC-I', 'HSC-Z', 'HSC-Y'))
+        else:
+            load_and_save_tract(args.repo, tract, filename, verbose=args.verbose)
```
wmwv: I suggest making this more directly obvious that it's controlling the filters.

```
if args.hsc:
   filters = ('HSC-U', 'HSC-G', 'HSC-R', 'HSC-I', 'HSC-Z', 'HSC-Y')
else:
   filters = ('u', 'g', 'r', 'i', 'z', 'y')

load_and_save_tract(args.repo, tract, filename, verbose=args.verbose, filters=filters)
```

I appreciate that you likely structured it the way you did above to avoid having to make a statement about the filters in the generic case.  But I think it's fine to assume the ugrizy filters as a default at this level.

Eventually we can make this a bit smarter by getting the filter list from the repo.  We can add that logic to this top-level main and then call `load_and_save_tract` with the filter list.  There won't be a need to call anything special options, then.

RobertLuptonTheGood:
I'm not sure exactly what you're doing, but the HSC names like "HSC-R" are aliased to e.g. "r".  That won't help if you're dealing with filenames.

scripts/merge_tract_cat.py
EiffL: Ok, sure, no problem

EiffL:
I did try to query the butler with simple filter names like "r" but I get this messages:
```
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 8523, 'patch': '0,5', 'filter': 'r'}, tag=set())
...
``` 
Works fine with `HSC-R`

jchiang87:
That may be due to my only downloading the relevant-looking bits of the repo.  When I use `r` at lsst-dev, the butler seems to find the data.

EiffL:
Ok, then, this may not be needed at all then :-) @jchiang87 if you don't need it, I'll just close this PR, it's a lot more convenient down the road if we use the aliases

jchiang87:
So what worked for me is this:
```
catalog = butler.get('deepCoadd_ref', dataId=dict(tract=8523, patch='3,3', filter='i'))
```
but when I try to run your script without the `--hsc` option, I do indeed get this:
```
python merge_tract_cat.py /datasets/hsc/repo/rerun/DM-13666/WIDE 8525 --verbose
CameraMapper INFO: Loading exposure registry from /datasets/hsc/repo/registry.sqlite3
CameraMapper INFO: Loading calib registry from /datasets/hsc/repo/CALIB/calibRegistry.sqlite3
CameraMapper INFO: Loading calib registry from /datasets/hsc/repo/CALIB/calibRegistry.sqlite3
Processing tract 8525, patch 0,0
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 8525, 'patch': '0,0', 'filter': 'u'}, tag=set())
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 8525, 'patch': '0,0', 'filter': 'g'}, tag=set())
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 8525, 'patch': '0,0', 'filter': 'r'}, tag=set())
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 8525, 'patch': '0,0', 'filter': 'i'}, tag=set())
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 8525, 'patch': '0,0', 'filter': 'z'}, tag=set())
  No locations for get: datasetType:deepCoadd_forced_src dataId:DataId(initialdata={'tract': 8525, 'patch': '0,0', 'filter': 'y'}, tag=set())
```

EiffL:
So, the alias is only defined in the reference catalog, not in the forced source catalog :-/ Ok, I'll fix that in the merge script

EiffL:
Oh, no, it's just that the filter option as no effect on the reference catalog. I added a dictionary to use the HSC names when querying the butler on the forced src, and use the simple filter names in the output catalog

EiffL:
@wmwv I've rebased the code, should be good to merge. Note that instead of re-running this script with the new convention for the filter names, I've updated the HDF5 already produced by myself and @jchiang87 with a very simple script stored here: `/global/cscratch1/sd/desc/hsc/dumb_pandas_rename_script.py` . In the future however, this would not be necessary, we can directly use the `merge_tract_cat.py` output

PR #189, opened 2018-07-10
Fixes the number of patches for HSC

EiffL:
@wmwv @yymao This is a very small PR that sets the number of patches for HSC. Unfortunately, the Butler doesn't let you query the number of patches, so you would have to dig into the repository tree to figure it out programmatically (and it quickly gets ugly when you have to handle different filter names).

We will have to regenerate the HDF5 for hsc with this new version of the script.

### DIFF
--- a/scripts/merge_tract_cat.py
+++ b/scripts/merge_tract_cat.py
@@ -82,8 +82,13 @@ def load_and_save_tract(repo, tract, filename, key_prefix='coadd', patches=None,
         Overwrite an existing HDF file.
     """
     butler = Butler(repo)
+
     if patches is None:
-        patches = ['%d,%d' % (i, j) for i in range(8) for j in range(8)]
+        # Extract the patches for this tract from the skymap
+        skymap = butler.get(datasetType='deepCoadd_skyMap')
+        numpatches = skymap.generateTract(tract).getNumPatches()
+        nx, ny = numpatches.getX(), numpatches.getY()
+        patches = ['%d,%d' % (i, j) for i in range(nx) for j in range(ny)]
 
     for patch in patches:
         if verbose:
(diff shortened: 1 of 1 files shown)


jchiang87:
@EiffL You can also just iterate via `for patch in tract` and use a try/except block and catch the `NoResults` exception to avoid assuming a specific number of patches per tract:
https://github.com/LSSTDESC/DC2_Repo/blob/u/jchiang/cmu_tutorials/Notebooks/DC2_Coadd_Run1.1p_Butler_Access_HSC_selection.ipynb

wmwv:
Or if you want to be explicit, you can
```
skymap = butler.get(datasetType='deepCoadd_skyMap')
tractId = 4850
tract = skymap.generateTract(tractId)
numpatches = tId.getNumPatches()
nx, ny = numpatches.getX(), numpatches.getY()
patches = ['%d,%d' % (i, j) for i in range(nx)
                            for j in range(ny)]
```

wmwv:
But I really like @jchiang87 's suggestion.  Replace the logic in `load_tract`:

EiffL:
ok great :-) I like this a lot! Replacing it now

EiffL:
Looks good, and it's running, I'm reprocessing tract 8523, will take a number of hours

yymao:
@EiffL The code should work but I thought the idea is to remove Lines 86 to 91 and change Line 93 to `for patch in tract:`?

EiffL:
Yes, my bad. Here, it's fixed :-) And no, I don't think we want to directly use `for patch in tract` at the next line, because we might want to specify a subset of patches

yymao:
Should edit `load_tract` accordingly too.

EiffL:
facepalm, thanks Yao! I think I should take a break from touching critical code for today.

EiffL:
Ok, everything seems to work.

PR #192, opened 2018-07-11
Issues/181 - Postage Stamp PSFs

wmwv:
@EiffL I've added a demonstration of getting the PSF at a location and an overly simplistic example of subtracting a scaled PSF from a postage-stamp image.


EiffL:
I'm looking at your tutorial, it's really good ! But what happened to the interactive plots where you could see the pixel and mask values ? Is it just a matter of choosing the backend ?

wmwv:
@EiffL Yes, it's the backend.  JupyterHub doesn't support the interactive backend `%matplotlib notebook` that allows the interactive plots because it doesn't support execution of arbitrary Javascript (this is an intentional feature).  JupyterLab does allow the interactive backend.

So the default is that it's set to `%matplotlib inline` so that it runs on JupyterHub, but it's set up such that if you're running under JupyterLab you can easily use `%matplotlib notebook` instead and the specified figure/frame numbers will keep the interactive plots on the right figures.

wmwv:
@RobertLuptonTheGood Thanks for the suggestion.  `ipympl` is not available in our installs on NERSC.  I'll look into this a bit more and see if we should request it to be added.

EiffL:
Well, actually, it works just fine for me using %matplotlib notebook under the default jupyter-dev hub at nersc... or am I missing something?
image

RobertLuptonTheGood:
I don't think `%matplotlib notebook` works in JupyterLab (it's fine in Jupyter)

wmwv:
@EiffL @drphilmarshall I've expanded and improved the documentation of the Jupyter vs. JupyterLab issue for the interactive backend.

Would one of you be kind enough to formally approve this PR so that I can merge?

Issue #183, opened 2018-07-06
Tutorial notebook(s) for extragalactic catalogs


yymao:
Add tutorial notebook(s) for extragalactic catalogs. The plan is to reproduce some DESCQA tests in the form of jupyter notebook.

yymao:
I started to add some notebooks in `u/yymao/cmu_tutorials` branch. See https://github.com/LSSTDESC/DC2_Repo/tree/u/yymao/cmu_tutorials/Notebooks

evevkovacs:
Submitted a PR last night and updated the notebook today with some additional comments. It is ready for review.

PR #194, opened 2018-07-12
N_vs_z distribution culled from descqa test

evevkovacs:
Here is a first pass example of a notebook culled from a descqa test. It has some more advanced features such as using the catalog iterator and accumulating histograms on the fly. It could use some more comments, but is good enough to pass to the reviewers for comments.


yymao:
This PR addresses #183.

wmwv:
@evevkovacs Could you kindly formally assign a reviewer?

patricialarsen:
@evevkovacs This looks good, it definitely works and it's pretty clear what it's doing. A couple of minor points but it's all preference rather than necessity:
- it would be good if you could make the place where you should input the binning in redshift more obvious, it took me a minute to work this out as you're using the function defaults currently. 
- you could link this to the default protoDC2 rather than your local version of a specific catalog to avoid confusion

evevkovacs:
Just committed a version that addresses Patricia's comments. Choice of Nzbins is made explicit, code defults to use protoDC2 with special access using private reader left in the comments to provide help for advanced users.

wmwv:
@evevkovacs This is a nice straight-forward example of a thing everyone will want to do and how to use iteration and filtering in the context of the GCR.

wmwv:
@evevkovacs Can you confirm that the intent is for the bright magnitude limit to always be 19?

Another possibility would be to show specific magnitude ranges.  I'm not familiar enough with the goal of the N(z) plots to understand which is more common/useful.

evevkovacs:
Yes, that was the intent for this example, bu tof course, it doesn't have to be 19. You can set it to whatever value you want. You can also make it [(m-1) for m in mag_lo] (or whatever), so the plots will show the distributions in magnitude bins, as you suggested above. (This could be left as an exercise for the reader) Both variants are employed in the DESCQA test. The driver is the data that is available, and both variants exist in the literature.

wmwv:
Great, thanks.

wmwv:
Doing magnitude bins is a great exercise for the reader.

wmwv:
@evevkovacs I've made some simplifications to the plotting logic that I hope will improve the readability for the user.

wmwv:
@yymao Are you happy with this Notebook now?

yymao:
Looks good, thanks @wmwv @evevkovacs .

evevkovacs:
Yes, I am happy. Just did a final test. Thanks @yymao and @wmwv  for the improvements in comments and presentation.

Issue #184, opened 2018-07-06
Tutorial notebook(s) for truth catalog


yymao:
Add tutorial notebook(s) for the truth catalog. The tutorial should use the reader, which does not yet exist (see https://github.com/LSSTDESC/gcr-catalogs/issues/130). 

We should also replace the reference catalog used in @slosar and students' matching notebook with the truth catalog (with GCR this should be just changing one line).

danielsf:
I'm almost ready to incorporate the truth catalog for static sources into the GCR framework.  These are the columns I imagine being included in the truth catalog.  Please pipe-up if I am neglecting something you think is important (@slosar @wmwv @cwwalter )

ra, dec -- in degrees

healpix_id  -- the index of the n_side=2048 nested healpixel containing the source

redshift

is_star -- an int (either 1 or 0)

has_agn -- an int (either 1 or 0)

is_sprinkled -- an int (either 1 or 0); objects with is_sprinkled==1 will have no counterpart in the extragalactic catalog

ugrizy -- the magnitudes of the source in all 6 LSST filters.  These magnitudes will be the mean magnitudes for variable sources.  For galaxies, they will include all components (bulge, disk, AGN).  No dust extinction will be applied, since Run 1.1p was run with a version of PhoSim that had accidentally turned off all dust.

danielsf:
also: galaxy_id, which I will set to -1 for stars

danielsf:
While our Milky Way model does have radial velocity, that information never makes it to ImSim/PhoSim, so the stars are at `redshift==0`

The sprinkler does assign a new `galaxy_id` for the purposes of generating the object identifier that gets passed to PhoSim, so `galaxy_id>0` for sprinkled objects.

Also: I remembered that the stars also have a unique identifier (`simobjid`) within the CatSim model, so that column will be `object_id`, which will be `==galaxy_id` in the case of galaxies and `==simobjid` in the case of stars.  Users will be have to use the `star` and `sprinkled` columns to figure out what ID they are actually looking at.

danielsf:
PR #198 opened for anyone interested

yymao:
Now that we have truth catalogs, should we update the matching notebook(s) to use truth instead of reference? What do you think, @danielsf @wmwv @slosar?

wmwv:
Yes

danielsf:
[...]

PR #198, opened 2018-07-16
U/danielsf/example/notebooks/truth

danielsf:
This notebook should address Issue #184 

It needs to be run against master of `yymao/generic-catalog-reader` and `gcer-catalogs`.  The first cell includes some code to redirect your environment to load clones of those packages.

Let me know if you think there is any other functionality that should be demonstrated.

### DIFF
--- a/Notebooks/DC2_truth_example.ipynb
+++ b/Notebooks/DC2_truth_example.ipynb
@@ -20,11 +20,22 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


yymao:
@danielsf 
- I've checked the notebook works with current `desc-python` kernel without the first cell, so the first cell can be removed.
- need to add title, notebook owner, last run data, and learning objectives.
- does SQL supports python-like syntax (e.g. `star==1`)? If so, change `star=1` to `star==1` in native filters. If not, the notebook should explain this and warn users the syntax difference.

danielsf:
This is ready to be re-reviewed

PR #201, opened 2018-07-18
Replace reference catalog with truth catalog

yymao:
This PR replaces the reference catalog with the truth catalog in the [FoF matching notebook](https://github.com/LSSTDESC/DC2_Repo/blob/u/yymao/use-truth-for-matching/Notebooks/FoFCatalogMatching%20Histogram.ipynb), and also removes the reference catalog notebook (there's not much content there anyway, you can see the original notebook [here](https://github.com/LSSTDESC/DC2_Repo/blob/master/Notebooks/DC2%20Reference%20Catalog%20Reader.ipynb)). 

This PR is ready to be reviewed.


Notebooks/FoFCatalogMatching Histogram.ipynb
```diff
@@ -1,296 +1,283 @@
...
    "outputs": [],
    "source": [
-    "from collections import defaultdict\n",
-    "import numpy as np\n",
-    "from astropy.coordinates import SkyCoord\n",
-    "import FoFCatalogMatching\n",
-    "from GCR import GCRQuery\n",
-    "import GCRCatalogs"
+    "import sys\n",
+    "sys.path.insert(0, '../../gcr-catalogs/')"
```
wmwv: What's the timescale for updating the default gcr-catalogs on NERSC to fix the need for this?
This specifically only works if someone has cloned the `gcr-catalogs` repo next to the `DC2_Repo`.

yymao:
Thanks, @wmwv. I made some minor updates. I'll merge this now. 

As for gcr-catalogs version, we'll certain install the master again soon. But think about it, I guess it's probably a good idea to teach people how to use master, as people may need to fix things in GCRCatalogs when they try these notebooks.

PR #202, opened 2018-07-18
U/dkorytov/cmu tutorial

dkorytov:
A simple cluster member cutout example and a cluster color-z distribution example notebooks.


dkorytov:
Reformated & clean up. 

I've added Patricia Larsen as an owner to the cluster cut out notebook since she wrote the text and expanded it. I couldn't find her user name in the repo--I can ask her tomorrow morning.

dkorytov:
,,,
label='Galaxy Members [{}]'.format(len(members))
,,,

doesn't work. It looks like GCRQuery returns a dictionary, so```len()``` returns the number columns. Maybe casting the dictionary into an astropy table, is an elegant solution.

yymao:
ah you're right. EIther option is fine by me.

dkorytov:
All implemented. 

For the second cluster cutout plotting, I wanted to show how to use numpy array selection. But people should already be familiar with that usage pattern.

yymao:
Thanks, @dkorytov. I made a minor fix to the astropy.table import statement.

PR #214, opened 2018-07-20
Create stamps in expected figure cell when using notebook backend

bechtol:
When running the `DC2_Postage_Stamps.ipynb` directly in Jupyter with the notebook backend, the `make_cutout_image` now creates a designated figure for the postage stamp instead of putting the image in whatever figure cell was last active.


kadrlica:
What about doing this in the cell where the function is called rather than in the function itself? That way the user has more control over the figure that the cutout is drawn to.

kadrlica:
Would this do it?
```
# Plot just one
i = 3
first = id_ra_dec[i]
ra, dec = first['RA'], first['DEC']
plt.figure('stamp')
cutout = make_cutout_image(butler, ra, dec, filter=filter, label="Object ID: %d" % id_ra_dec[i]['ID'])
```

bechtol:
Following suggestion by @kadrlica . 

I think the part that is rather non-intuitive to me is that even when the `show` argument is set to `False` in `make_cutout_image`, figures in other cells are affected, and this leads to particularly strange behavior when cells are run out of the linear sequence. I think the suggested change partially improves the situation.

wmwv:
In brief, matplotlib's global state is really a pain.

* I've added a `frame=` keyword to `display_cutout_image` so that the `afwDisplay.Display` call inside gets the right frame.
* I've explicitly specified the frame for every call that involves plotting
* I've explained that one will get a figure even with `show=False` in the `make_cutout_image` call in interactive mode, because that automatically triggers display of the plot with no `plt.show` required.

wmwv:
@kadrlica Could you take a look at my changes to see if they make sense to you?

kadrlica:
Do we want to comment on why the `i = 1` and `i = 2` cutouts looks like this?

image

Here's an associated 3k x 3k cutout and mask:

image
image

I don't know what's going on here, but looks like some interpolation over missing data. Maybe we leave it as an exercise to the user to figure out how to use the mask plane and flags to identify objects of this type?

kadrlica:
And now the source code manifests itself...
https://github.com/lsst/display_matplotlib/blob/master/python/lsst/display/matplotlib/matplotlib.py

All shall be revealed! [For example](https://github.com/lsst/display_matplotlib/blob/d076f1fb4d29987d637edbb2680bc64a89bab192/python/lsst/display/matplotlib/matplotlib.py#L290-L293):
```
        mappable = ax.imshow(dataArr, origin='lower', interpolation='nearest',
                             extent=(bbox.getBeginX() - 0.5, bbox.getEndX() - 0.5,
                                     bbox.getBeginY() - 0.5, bbox.getEndY() - 0.5),
                             cmap=cmap, norm=norm)
```

I'm beginning to think that my dive into the `afw.display.matplotlib` backend may merit it's own discussion thread (or notebook).

Issue #215, opened 2018-07-20
Review of GCR Access.ipynb


chihway:
This issue is to track feedback on the notebook [GCR Access.ipynb](https://github.com/LSSTDESC/DC2_Repo/blob/master/Notebooks/DC2%20Coadd%20Run1.1p%20GCR%20tutorial%20--%20Part%20I:%20GCR%20Access.ipynb). It runs through smoothly, so just a few suggestions (none of these are urgent so please take or leave whatever you think appropriate):

* link to what CModel and HSM are?
* is there a function (within GCR or otherwise) to convert between RA/DEC and track number/patch number? that seems useful to have and show here
* explain a little more what "native filter" means, is that ONLY for the tracks or are there other things in this category we can filter?
* I don't fully understand what the "return_iterator=True" is doing and how it could help in doing analyses, perhaps explain a little?
* in the comparison of full vs. cleaned objects for DC2, maybe print out the number of objects? just to convince people that you are cutting out something

A general question:
* would you recommend running the notebooks in the jupyter lab environment or normal notebook mode?

fjaviersanchez:
@chihway Thanks for your comments. I have tried to address all of them in the PR above. 

For your general question, I haven't put anything on the notebook because I think I may confuse the users. I have tested them in both jupyter-dev and JupyterLab, and they work similarly (I have found that the kernel dies more often in JupyterLab for some reason but it might just be a temporary thing). There is added functionality if you use JupyterLab (see [here](http://jupyterlab.readthedocs.io/en/latest/user/interface.html)) but I don't have enough experience to recommend one over the other. Maybe @EiffL or @yymao have a stronger opinion on this.

chihway:
Looks good, thanks @fjaviersanchez !

yymao:
fixed by #218

PR #218, opened 2018-07-20
Addressed comments from Chang, Bogart and Mudur

fjaviersanchez:
This PR tries to address the comments for the Notebook `DC2 Coadd Run1.1p GCR tutorial Part I: GCR Access.ipynb` from Chihway and others in #215


yymao:
@EiffL @fjaviersanchez I updated the native filter description. Please take a look. Note the native filters don't work on *any* native quantities; it only works on quantities that are related to the underlying data format.

PR #223, opened 2018-07-24
match truth catalog to extragalactic catalog

danielsf:
This PR adds cells to the FoF matching notebook that show how to cross-reference the truth catalog with the extra-galactic catalog.


yymao:
@danielsf thanks for adding this. I switched to pandas for the ID matching step. I thought that makes the code cleaner and easier to understand. Also updated the examples. What do you think?

PR #227, opened 2018-07-25
Improved challenge in gcr coadd tutorial

EiffL:
This PR turns removes/comment the code from cells in `DC2 Coadd Run1.1p GCR tutorial -- Part III: Guided Challenges.ipynb` and puts the solution in a hidden python script, so that people can fill in the blanks.
Notebook should still be able to run without errors but will not produce plots.


EiffL:
@yymao @drphilmarshall could we rerun beavis-ci to render this updated notebook ?

yymao:
I believe @drphilmarshall can trigger the job manually. We should wait for #228 being merged.

PR #239, opened 2018-07-31
incorporate WL group simulation needs in DC2 planning document

rmandelb:
While I was playing with the DC2-production repo and the new DC2-analysis repo, I discovered that I had some uncommitted changes to the DC2 planning document from a while ago.  This PR is the result of my committing those changes on a branch.  At some point there had been a question of the WL group DC2 simulation needs, which were driving some of the design choices and validation criteria in DESCQA; this additional paragraph includes those needs as originally articulated by the WL group [here](https://docs.google.com/presentation/d/1xHtwdI3C9c1cobiDFjt4qeZB45ojHn1QSNBrg2Uzx68/edit?usp=sharing).  Obviously the point is not to influence DC2 design at this stage, but rather to properly document why certain things were considered high priority.

I don't think there are any surprises here, since it's a reflection of what the WG has discussed multiple times.


rmandelb:
Thanks.  Since Tim is on semi-vacation and this is not controversial in nature, I'm going to go with one convener is enough.  Merging now.

Issue #237, opened 2018-07-27
Get Dask working at NERSC to analyze DC2


wmwv:
Get Dask working at NERSC to analyze DC2.

[x] Single-node
[ ] Job queue multi-node
http://dask-jobqueue.readthedocs.io/en/latest/configurations.html#nersc-cori

wmwv:
This work has ended up continuing in #190 
In brief, I have Dask working in single-node and job queue multi-node (using 'debug' queue) on Cori.  It works but is no faster.  Follow #190 for more.

heather999:
@wmwv Where is the notebook associated with these tests?  at this point, I just have some curiosity.  Was there any difference in using SCRATCH? I looked into the details of using the Burst Buffer, and plan to write those up. I imagine that may be generally useful outside of this specific work.

wmwv:
The Notebooks are in the respective branches:

https://github.com/LSSTDESC/DC2-production/blob/issues/190/Notebooks/DC2_Catalog_Performance.ipynb
https://github.com/LSSTDESC/DC2-production/blob/issues/237/Notebooks/object_catalog_performance_dask.ipynb

wmwv:
@heather999 #190 was just merged to master, and I renamed the Notebook above.  So you can look now in

https://github.com/LSSTDESC/DC2-production/blob/master/Notebooks/object_catalog_performance.ipynb

wmwv:
Closed by #276

PR #240, opened 2018-07-31
Issues/190 - Object Catalog Performance

wmwv:
1. Using a trimmed file consistent only of the columns exposed through the DPDD is 10 times smaller than the full file.
2. Analyses using GCRCatalogs are 10-30 faster with the trimmed files than with the full files.
3. Analyses with the HDF5 format='table' are 1-2 times slower than HDF5 format='fixed' files.
4. Directly accessing vs. using the GCR iterator had no significant performance impact.
5. Pandas takes 10 times longer to load the catalog than GCRCatalogs takes to initalize it. But once loaded, Pandas is 100,000 times faster than GCRCatalogs.
6. A preliminary Dask analyses is presented here, but a fuller development of Dask will be done in issues #237.

I'm persisting the output results in the cells because the performance testing takes several hours to run on Cori (and the results are lightweight text lines.)


wmwv:
I plan on 
1. renaming this Notebook, "object_catalog_performance.ipynb"
2. keeping it here in DC2-production

yymao:
@wmwv I am not surprised that pandas is faster than GCR but am very surprised that it is 100,000x faster. Looking into your notebook, I don't think that is a fair comparison. The time to compute mean color for GCR includes data load time, since you called `clear_cache` before the computation. While in the case of pandas, all data are already in memory `df`. 

A more fair comparison would be loading the catalogs first for some other use:
```python
data = gc.get_quantities(['ra', 'dec'])
```
and then, without clearing up the cache, time:
```python
%%timeit
compute_mean_color_faster(gc)
```

wmwv:
Oh, yes, thank you.  That makes much more sense!  I will do that comparison.

wmwv:
Hmmm... I did that comparison and got the same results.

```
# Give chance to fill cache
_ = gc_table_trim.get_quantities(['ra', 'dec'])
```

```
%%timeit
compute_mean_color_faster(gc_table_trim)
```
gives
```
25.7 s ± 1.43 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

wmwv:
And now I'm all confused and getting inconsistent results.  Give me a few minutes here.

wmwv:
Wow.  The surprising result above is due to the 'fixed' vs. 'table' format.

1. For the *full* catalog files in *fixed* format it's 781+-71 ms to do the mean color computation once the catalog is in cache.
2. For the *trim* catalog files in *fixed* format it's 366+-43 ms to do the average color computation.
3. For the *trim* catalog files in *table* format it's 24+-1 *second* to do the average color computation.

yymao:
With cache turned on, I'm guessing now you'll also see the difference between `return_iterator` set to `True` and `False`.

wmwv:
Using the iterator is about 30% faster.

wmwv:
@yymao Thanks again.  I improved the utility of the Notebook by presenting a better comparison of the performance of GCR with caching.

Take a look, make suggestions about other things I could do better, or approve if you're happy.

yymao:
@wmwv I pushed a commit here (but didn't rerun the notebook, so the timing numbers are *not* reflecting the updates). There are several points I'd like to mention:

- There's a bug in `load_all_tracts_into_pandas`. It was repeatedly loading only one tract rather than multiple tracts, which results in a much smaller dataframe. The bug is [fixed here](https://github.com/LSSTDESC/DC2-production/pull/240/commits/cc3ec8cf78b7cd14b29551d6b00c9fcc298c8b0a#diff-d125c6118200e8fe86f5d155c6289c3bR1114). When this is fixed, I found the time it takes to calculate mean(g-r) is about only 2x faster in pandas than in GCR+table format (about 3x faster than GCR+fixed). Also, the time for pandas to concat all pandas dataframes becomes much longer once this bug is fixed. 

- I added two `%%prun` cells. Profiling suggests that the GCR (i) didn't cache for table format and (ii) was overcautious with missing columns. Both are now fixed in https://github.com/LSSTDESC/gcr-catalogs/pull/167. Once those are fixed, profiling suggests that this comparison between pandas and GCR is in fact comparing how fast numpy vs. numexpr+bottleneck can calculate the mean. I think this is potentially misleading as it is not really about the underlying data access performance.

yymao:
Also, profiling also suggests that the GCR method is limited by the underlying data format. Most of the time GCR spends on is in fact generating NaN arrays for missing columns, because different tracts/patches do *not* have exactly the same set of columns. If we can fix this in the underlying data files, we can much speed up the reader performance.

wmwv:
@yymao I've worked a bit on the phrasing and summary statements.  Do you have any more wording and presentation changes you'd like?

PR #246, opened 2018-08-05
U/wmwv/validate Run 1.2p-test validate_drp results

wmwv:
@johannct @boutigny @fjaviersanchez @rmjarvis @katrinheitmann @cwwalter 

I've run `validate_drp` on the DC2 Run 1.2p-test data.

* I recommend proceeding to run the full coadds and associated forced photometry for Run 1.2p.
* I share @boutigny 's concern about the photometry.  It is worse than expected for reasons that aren't immediately obvious.
* Overall: The reproducibility of astrometry looks great (AMx); the photometry could be better (PAx); the ellipticity residual correlations look reasonable (TEx).

* These don't directly answer the question of "are the flats OK", but as discussed at some length elsewhere, that's a hard question to ask directly.  What we really can ask is "how well do the catalogs perform".  While the photometry repeatability numbers offer some source of concern, the fuller question is "is sufficient information available in the calibration products to recover performance and do we know how to use it?".  I think this is something we will answer through continued and iterative investigations based on the full set of coadd + calexp photometry.

* JSON files and plots in 
/global/projecta/projectdirs/lsst/global/in2p3/Run1.2p-test/w_2018_28/validate_drp  
See in particular the '*PA1.png' and '*photometry.png' plots.

* The scripts that generated these are added in this pull request.

* Technical Notes
This switch to `obs_lsstCam` required just a one-line change to validate_drp branch:
   tickets/DM-15323

* Questions:
1. Should these scripts live here in DC2-production/Validation?
    - I think these definitely live here in DC2-production because these are checks on the produced and reduced images through the catalogs.  Longer term, in my vision DC2-production generates a Data Release, including the documentation of its performance.

2. Are the instructions clear?
    - For this test it's perhaps not important that you can actually re-run these, just mostly that what was done was documented.  When we run `validate_drp` on a full set of Run 1.1 or Run 1.2, that should be reproducible.


yymao:
I ported these plots to the DESCQA web interface, just for the ease of browse them without needing to log in to cori: 

https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-08-05&test=validate_drp 
(All plots are listed on the left. Click them to enlarge. The panel divider is movable.)

(For now I did this manually, but if this is something people find useful, we can discuss how to do this automatically.)

fjaviersanchez:
@wmwv thanks a lot for these plots. I think that the issue with the photometric repeatability may be solved after @jchiang87's findings (we were using extended sources for calibration that weren't fully fitting in the default 12 arcsec aperture). This turned out to shift the median value for imSim's sources in ~40 mmags (for PhoSim the median was centered within the margins that we allowed for the extinction problems but the distributions looked fine). Maybe it's worth reprocessing a couple of these images with the new settings and check whether the photometric repeatability improves or not.

RobertLuptonTheGood:
You *never* want to use extended sources for photometric calibrations!  Not only is their profile unknown (so you can't apply aperture corrections rigorously), but the colour terms from the standard catalogue are less well defined than for stars as they have a wider range of SEDs.

fjaviersanchez:
@RobertLuptonTheGood thanks! I think there was a misconception among us while using the DM stack  that it automatically chose stars by default. @jchiang87 can elaborate more, but I think we were setting some flag to use unresolved objects, however, we weren't putting an upper limit on their extendedness. That's why we accidentally ended up using galaxies in the photometric calibrations. I also have to say that the ratio galaxies/stars included in the calibrations was low (well below 5% in the exposures that I checked) but the effect was noticeable enough in the imSim images that we ended up realizing that there was something going on.

RobertLuptonTheGood:
Well, we try to choose stars but in general that's not a binary decision.  There's code about to hit master that uses some ML to give a better split.  In general the input catalogue isn't all that clean (think SDSS or PanSTARRS) so you can't just rely on their classification.
Anyway, I'm glad that you sorted this out, and that 40mmag stuck out as a clear problem!  Of course, these days 40mmag isn't that much.

cwwalter:
Robert,

Are there settings other than these?

```
config.calibrate.photoCal.match.sourceSelection.doUnresolved=True
config.calibrate.photoCal.match.sourceSelection.unresolved.minimum=None
config.calibrate.photoCal.match.sourceSelection.unresolved.maximum=None
```

These were the defaults and they working working as Nones caused the doUnresolved flag to be ignored.  See https://lsstc.slack.com/archives/C932BQ550/p1533150355000356 for more details.

PR #260, opened 2018-09-05
initial guess at some package versions for Run 2.0i

jchiang87:


jbkalmbach:
`Twinkles` still requires an update once the final `cosmoDC2` galaxies are ready.

yymao:
@danielsf @jchiang87 are we still planning on going into production tomorrow? And if yes, should I tag `gcr-catalogs` even though there's still unfinished work, or are you planning to just use the master branch for instance catalog generation?

jchiang87:
We are planning to go into production tomorrow, but I'm not sure which `gcr-catalogs` will be, are being, or have been used.

danielsf:
InstanceCatalog production is going to start (if there is any justice in the Universe) in a few hours

we will just use master

yymao:
Okay, just note the commit hash that you use...

rmjarvis:
I'd recommend tagging with something like `lsst.dc2.0a` or something.  Makes it easier to find later rather than just some random SHA.  (Although you can tag after the fact if you start production with the current master at the time.)

yymao:
I've tagged `gcr-catalogs` at [v0.9.0](https://github.com/LSSTDESC/gcr-catalogs/releases/tag/v0.9.0) in case people want to use a tagged version for production.

wmwv:
@jchiang87 I propose closing this PR without further action.  Do you agree?

jchiang87:
sure.  closing now.

PR #268, opened 2018-10-01
Document use of merge_tract_cat.py in scripts/README.md

wmwv:
Documentation of using merged_tract_cat.py to generate the Run 1.1 Object summary files.

The same should work for Run 1.2 with very minor changes, suggested by this documentation.


JoanneBogart:
I had no trouble understanding it as it stands, but suggest that all command excerpts be delimited by ``` ...  ``` to preserve formatting.   This is done for some of them, but not for the first two.
The content is fine.

wmwv:
@JoanneBogart Thanks.  Fixed.

PR #269, opened 2018-10-03
U/wmwv/hdf to parquet

wmwv:
Adds `convert_merged_tract_to_dpdd.py` that produces FITS, HDF5, and Parquet files that have column names following the DPDD.  There is one file per tract+patch.  File names start with `dpdd_`.

Output files on NERSC in

/global/projecta/projectdirs/lsst/global/in2p3/Run1.1/object_catalog/


wmwv:
@JulienPeloton Could you review for utility for your specific use of case of testing for Spark?
@JoanneBogart Could you review for future QSERV ingestion?
@yymao Could you review for general code style, efficiency, clarity, and use of the GCR?

scripts/convert_merged_tract_to_dpdd.py
```diff
@@ -0,0 +1,129 @@
...
+    This is done in one function because we're doing this chunk-by-chunk.
+    """
+    columns = cat.list_all_quantities()
+    columns.extend(['tract', 'patch'])
+
+    # tract is an int
+    # but patch is a string (e.g., '01' for '0,1')
+    outfile_base_format = '{base}_tract_{tract:04d}_patch_{patch:s}'
+    quantities = cat.get_quantities(columns, return_iterator=True)
+    for q in quantities:
```
yymao: This line at face face seems to be iterating over quantities, but it's actually iterating over chunks (patches, in this case). I suggest be explicit about this in the variable name:

```python
for quantities_this_patch in quantities:
```

scripts/convert_merged_tract_to_dpdd.py
```diff
@@ -0,0 +1,129 @@
...
+        if 'tract' not in q or len(q) < 1:
```
yymao: Would this ever happen?  I don't think so.

scripts/convert_merged_tract_to_dpdd.py
```diff
@@ -0,0 +1,129 @@
...
+            continue
+
+        df = pd.DataFrame.from_dict(q)
+
```
yymao: Once `q` (or `quantities_this_patch` if we rename it) is converted to `df`, it is no longer needed, and can be deleted so that garbage collector can collect it when necessary. 
```python
del q
# OR
del quantities_this_patch
```

scripts/convert_merged_tract_to_dpdd.py
```diff
@@ -0,0 +1,129 @@
...
+        # We we know that our GCR reader will chunk by tract+patch
+        # So we take the tract and patch in the first entry
+        # as the identifying tract, patch for all.
+        tract, patch = q['tract'][0], q['patch'][0]
```
yymao: Since `q` is deleted, here we need to do
```python
tract, patch = df['tract'][0], df['patch'][0]
```

scripts/convert_merged_tract_to_dpdd.py
```diff
@@ -0,0 +1,129 @@
...
+        patch = ''.join(patch.split(','))
```
yymao: Took me a second to realize this is just to remove the comma. We can be more explicit about the intention:   
```python
patch = patch.replace(',', '')
```

scripts/convert_merged_tract_to_dpdd.py
```diff
@@ -0,0 +1,129 @@
...
+        key = '{key_prefix:s}_{tract:04d}_{patch:s}'.format(**info)
+        df.to_hdf(outfile_base+'.hdf5', key=key)
+
+        df.to_parquet(outfile_base+'.parquet',
+                      engine='fastparquet',
+                      compression='gzip')
+
+        if verbose:
+            print("Writing FITS DPDD file for ", tract, patch)
+        Table.from_pandas(df).write(outfile_base+'.fits')
```
yymao: I would expand this to:
```python
df = Table.from_pandas(df)
df.write(outfile_base+'.fits')
del df
```
This again is to help gc to collect when necessary.

scripts/convert_merged_tract_to_dpdd.py
yymao: Once you obtain the tract and patch, do you still want to store them in the output files? If not, you can delete them from the data frame
```python
df.drop(columns=['tract', 'patch'], inplace=True)
```

scripts/convert_merged_tract_to_dpdd.py
wmwv: Yes, it did.  That's why I added that check.  I got an error message about `tract` not being found.

But really I bet that the check just needs to be for `len(q) < 1`.

scripts/convert_merged_tract_to_dpdd.py
wmwv: Why does this help the garbage collector?

The command explicitly doesn't bind a name to `Table.from_pandas(df)`.  The gc should thus already know it can remove the object.

scripts/convert_merged_tract_to_dpdd.py
wmwv: I'm waffling on this.  I think it will be useful practically, but it also bugs me that it injects non-DPDD column names.

scripts/convert_merged_tract_to_dpdd.py
yymao: The pandas data frame is no longer needed once converted to astropy table. Current implementation keeps both copies during writing.

yymao: I don't think the memory usage actually matters in our case, but just think it's generally a good idea to clean up since we are dealing with very large tables.

wmwv: Aside from the factor of 2, the `gc` should definitely know it can clean up with the next loop.

scripts/convert_merged_tract_to_dpdd.py
wmwv: And now I can't reproduce the error...

yymao: I checked GCR source code again and still didn't think the reader will ever return an empty dictionary. 

I suggest removing this check. If an empty dictionary is returned, it's useful to get an exception because it is not expected.

wmwv: Thanks for checking.  Check removed.

scripts/convert_merged_tract_to_dpdd.py
```diff
@@ -0,0 +1,135 @@
...
+def convert_cat_to_dpdd(cat, reader='dc2_coadd_run1.1p',
+                        **kwargs):
+    """Save DPDD-named columns files for all tracts, patches from input GCR catalog.
+    """
+    columns = cat.list_all_quantities()
+    columns.extend(['tract', 'patch'])
+
+    quantities = cat.get_quantities(columns, return_iterator=True)
+    for quantities_this_patch in quantities:
+        write_quantities_to_files(quantities_this_patch, **kwargs)
```
yymao: This is more a personal preference, but I'd convert the dict to pandas dataframe here.
```python
for quantities_this_patch in quantities:
    quantities_this_patch = pd.DataFrame.from_dict(quantities_this_patch)
    write_quantities_to_files(quantities_this_patch, **kwargs)
```
And let `write_quantities_to_files` takes dataframe directly. 

This way, `quantities_this_patch` won't take up twice the memory.

scripts/convert_merged_tract_to_dpdd.py
```diff
@@ -0,0 +1,135 @@
...
+ 
+def write_quantities_to_files(quantities_this_patch,
+                              hdf_key_prefix='object',
+                              parquet_compression='gzip',
+                              verbose=True,
+                              **kwargs):
+    """Write out quantitites to HDF, FITs, and Parquet files."""
+    df = pd.DataFrame.from_dict(quantities_this_patch)
```
yymao: continuing my comment above
```python
def write_quantities_to_files(df,
                              hdf_key_prefix='object',
                              parquet_compression='gzip',
                              verbose=True,
                              **kwargs):
    """Write out quantities to HDF, FITs, and Parquet files."""
```
(and remove line 66) (note the typo "quantities" in the docstring)

wmwv:
Merged to master.

Thanks, everyone.

PR #276, opened 2018-10-11
Issues/237 - Explore basic performance with Dask

wmwv:
This PR adds a Notebook with some preliminary explorations of using Dask to directly read the trimmed HDF5 files for the Object Table.


SimonKrughoff:
See [this notebook](https://github.com/lsst-sqre/notebook-demo/blob/master/bokeh-WIP/gapminder.ipynb) for an example on how to inline bokeh.

wmwv:
@JulienPeloton 
It does suppress the warnings for me in either `Python [default]` or the custom `desc-python-dask` kernels.

I can't figure out if PEP-565, which was implemented in Python 3.7 changed the default behavior.
 https://www.python.org/dev/peps/pep-0565/

Could you test under a Python 3.7 environment?  That's what we're formally supporting.

JulienPeloton:
@wmwv OK thanks it seems to work under a Python 3.7.

However, NERSC brings support only for python 3.5/6, not 3.7:

```
peloton@cori06:~$ module avail python

---- /usr/common/software/modulefiles ----
python/2.7-anaconda              python/2.7-anaconda-4.4(default) python/2.7-anaconda-5.2
python/3.5-anaconda              python/3.6-anaconda-4.4          python/3.6-anaconda-5.2
```

So is LSST/DESC using a user-defined python installation that I need to load?

wmwv:
Yes, see

https://confluence.slac.stanford.edu/display/LSSTDESC/Using+Jupyter-dev+at+NERSC

wmwv:
And we're not trying to break anything under Python 3.6, but given that this is mostly meant as a demonstration Notebook, I think the present warning issues being solved in Python 3.7 but just producing more output in Python 3.6 is fine.

wmwv:
@JulienPeloton You're totally right!  Those are Python 3.6 kernels!
Hmm... I guess I just had something wrong in my memory of what version those builds were on.

JulienPeloton:
OK so all good to me.  Just one small thing, the Notebook is 19 MB. Is there a way to reduce its size?

wmwv:
I don't know how to reduce the size while preserving the plots.  I'd be happy to take suggestions.

I'm intentionally keeping the outputs for these performance measurement Notebooks because the tests generally take so long to run (~hour).

PR #287, opened 2018-10-18
Issues/285

wmwv:
The Parquet format allows Dask to perform *much* better than HD5.  3-7 seconds instead of 70 seconds.

[Edit: 0.7 seconds for Parquet simple single-file format (~4 GB file).]


cwwalter:
That's really interesting. I wonder if it is something intrinsic or if it has to do with how we structure and/or open the HDF5 files?

wmwv:
I think it's intrinsic.   Selecting only two columns from a column-based storage format should be much faster.  The HDF5 files are stored by row.

In principle, of course, HDF5 is a container and could store things in arbitrary format.  In practice, I mean what you get when you say `df.to_hdf(...)`, which is a row-based representation of the Pandas dataframe.

cwwalter:
> I think it's intrinsic. Selecting only two columns from a column-based storage format should be much faster. The HDF5 files are stored by row.

I see. That's independent of whether you use the 'table' format  or index data columns?

cwwalter:
And so the hypothesis now is that the main issue here is that you are selecting a sub-set of column and if you were using (say) all of them we might not see such a big difference but they would be be faster than doing a single serial operation?

JulienPeloton:
Note however that the size on disk of the datasets varies: 4.5 GB for Parquet Hive, 3.6 GB for Parquet simple, and 6.7 GB for HDF5. So part of the difference in execution time when loading comes from here I think.

wmwv:
I've emphasized the excellent performance of the 'simple' single-file Parquet format and added a dask.distributed run with the simple format.

JulienPeloton:
Excellent, very comprehensive 👍

PR #295, opened 2018-11-15
Add --name option to merge_tract_cat.py.  Default to 'object_'

wmwv:
Files will now be named, e.g., 'object_tract_4850.hdf5'
instead of 'merged_tract_4850.hdf5'


wmwv:
This is the correct PR for this very small change.

PR #300, opened 2018-11-21
start debugging Run1.2p

plaszczy:
Although the final dataset is likely to change this is a notebook useful for any prod


plaszczy:
thanks @JulienPeloton and good points.
I don't understand your comment about the 20 tracts. what do you ant to chect? Concerning the colormap of course I started with viridis but need to switch to this one to explicit bad behavior. If you know other color-blind tables we can do the test.

plaszczy:
right for tracts (so I should move the ref to the skymap before)
it coudl be nice to draw the outlines of the tracts from the minmax(ra,dec) per tract byt maybe painfull to overplot.

johannct:
If you look for instance at http://hsca.ipmu.jp/public/scripts/showVisitSkyMap.html you may see how to plot tracts or patches based on their innerBBox property

JulienPeloton:
@johannct nice, thanks!

plaszczy:
right but not doable, the butler runs in a dedicated  shifter image

PR #301, opened 2018-11-24
improve convert_merged_tract_to_dpdd.py

### LINKED ISSUE #293: add dedicated scripts to write FITS/Parquet objects catalog output
currently convert_merged_tract_to_dpdd.py (tries to) write 3 data format outputs: HDF5, FITS, Parquet.
This is unstable. The fits files are written by patch, which makes them useless, and the python parquet writing is fragile and fails sometimes.
I propose that convert_merged_tract_to_dpdd only deals with HDF5 output (which is robust) and FITS/parquet would come in a later stage that reads the former. This allows to write FITS files per tracts and solve the parquet instability.

yymao:
Prompted by @plaszczy's question, I updated `convert_merged_tract_to_dpdd.py` to accommodate the new object catalogs (1.2p/i). In particular, this PR:
- Merges the three functions `convert_all_to_dpdd`, `convert_tract_to_dpdd`, and `convert_cat_to_dpdd` into just one (`convert_cat_to_dpdd`).
- Makes a new option `reader_config_overwrite`, instead of hard coding it. 
- Reads in catalog config to get the file pattern, instead of hard coding it. 
- Generalizes the code so single-tract and all-tract workflows are the same.
- Uses `reader_config_overwrite` to set `use_cache`.
- Updates "coadd" -> "object".
- Ensures no repeated columns.
- Minor code style updates and typo fixes.
- [added 11/30] Add options to write out only specified format(s) [fixes #293]

### DIFF
--- a/scripts/convert_merged_tract_to_dpdd.py
+++ b/scripts/convert_merged_tract_to_dpdd.py
@@ -62,8 +62,11 @@ def convert_cat_to_dpdd(reader='dc2_object_run1.1p',
 def write_dataframe_to_files(
         df,
         filename_prefix='dpdd_object',
+        generate_hdf=True,
+        generate_fits=True,
+        generate_parquet=True,
         hdf_key_prefix='object',
-        parquet_scheme='hive',
+        parquet_scheme='simple',
         parquet_engine='fastparquet',
         parquet_compression='gzip',
         append=True,
@@ -79,6 +82,12 @@ def write_dataframe_to_files(
         Pandas DataFrame with the input catalog data to write out.
     filename_prefix : str, optional
         Prefix to be added to the output filename. Default is 'dpdd_object'.
+    generate_hdf : bool, optional (Default: True)
+        Whether or not to generate HDF output
+    generate_fits : bool, optional (Default: True)
+        Whether or not to generate FITS output
+    generate_parquet : bool, optional (Default: True)
+        Whether or not to generate Parquet output
     hdf_key_prefix : str, optional
         Group name within the output HDF5 file. Default is 'object'.
     parquet_scheme : str, optional   ['simple' or 'hive']
@@ -102,7 +111,7 @@ def write_dataframe_to_files(
     tract, patch = df['tract'][0], df['patch'][0]
     patch = patch.replace(',', '')  # Convert '0,1'->'01'
 
-    # Normalise output filename
+    # Normalize output filename
     outfile_base_tract_format = '{base}_tract_{tract:04d}'
     outfile_base_tract_patch_format = \
         '{base}_tract_{tract:04d}_patch_{patch:s}'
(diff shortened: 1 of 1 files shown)


plaszczy:
I am not sure using 'hive' as default parquet scheme is a good option. Looks like 'simple' is more efficient

yymao:
Changed `parquet_scheme` default to `simple`. Also, user-specified `parquet_scheme` did not propagate to the actual function call; this is now fixed. 

I also added options to disable the generation of certain types.

scripts/convert_merged_tract_to_dpdd.py
```diff
@@ -230,17 +193,20 @@ def write_dataframe_to_files(
...
+    parser.add_argument('--reader', default='dc2_object_run1.1p',
                         help='GCR reader to use. (default: %(default)s)')
-    parser.add_argument('--parquet_scheme', default='hive',
+    parser.add_argument('--no-hdf', action='store_false', dest='generate_hdf',
+                        help='disable HDF output')
+    parser.add_argument('--no-fits', action='store_false', dest='generate_fits',
+                        help='disable FITS output')
+    parser.add_argument('--no-parquet', action='store_false', dest='generate_parquet',
+                        help='disable Parquet output')
+    parser.add_argument('--parquet_scheme', default='simple',
```
wmwv: 1. I called it `--write_{fits,hdf,parquet}' and I would suggest that be the names of the options here.
2. Add both `--write-*` and `--no-*` command-line options.  Use `%(default)s` in the help string so that the help clearly prints out the default selected. 

E.g.,
```
    parser.add_argument('--write-hdf', default=False, action='store_true',
                        help='Write HDF output.  (default: %(default)s)')
    parser.add_argument('--write-fits', default=False, action='store_true',
                        help='Write FITS output.  (default: %(default)s)')
    parser.add_argument('--write-parquet', default=False, action='store_true',
                        help='Write Parquet output.  (default: %(default)s)')
   parser.add_argument('--no-hdf', action='store_false', dest='write-hdf',
                        help='Disable HDF output.  (default: %(default)s)')
    parser.add_argument('--no-fits', action='store_false, dest='write-fits',
                        help='Distable FITS output.  (default: %(default)s)')
    parser.add_argument('--no-parquet', action='store_false', dest='write-parquet',
                        help='Write Parquet output.  (default: %(default)s)')
```

yymao: @wmwv I'll be happy to add `--write-*` options but I want to know what's the expected default behavior when no options are given? 

If the default behavior (without any options) is to write out all formats, then I suggest that we change `--write-*` to `--write-*-only` so that these option actually has effects, and we can keep `--no-*` options too. [Note that in this case, the implementation in your example would not work as what we describe here, but I can implement it correctly.]

If the default behavior (without any options) is to write nothing at all, then I suggest that we add the `--write-*` options and also a `--write--all` option, but remove the `--no-*` options.

yymao: Alternatively, we can get rid of `--write-*` and `--no-*`, and just do this:

```python
parser.add_argument(
    '--format', 
    nargs='+',
    default=['all'],
    choices=['all', 'hdf', 'fits', 'parquet'],
)
```

wmwv:
My vote is for just Parquet to be the default.

I like the `—write` nargs option better, too, just as a general way to implement any of this.

yymao:
@wmwv OK, I am trying to interpret your comment above --- so you like (1) `write` option, (2) `nargs` method, and (3) default behavior to write only `parquet`. 

Combine these I think the implementation would be:

```python
parser.add_argument(
    '--write', 
    nargs='+',
    default=['parquet'],
    choices=['all', 'hdf', 'fits', 'parquet'],
)
```

PR #318, opened 2019-01-25
Issues/317

wmwv:
Using Pandas instead of AstroPy to join across filters speeds up AFW->HDF5 conversion by a factor of ~8 when testing on one patch (tract: 5066, patch='3,0').

A tract (5066) with 31 filled-in patches took 15 minutes instead of 2-4 hours.

The new method produces an HDF file of the exact same size as the previous now.  I'll do explicit validation over the next few days.


wmwv:
I've compared the catalogs generated by the old and new methods in tract 4850 and confirm that they are identical.

PR #321, opened 2019-02-07
Issues/315

wmwv:
Run 1.2p files prepared and installed.  DPDD versions prepared.  README updated.

I'm particularly interested in suggestions for making the README clearer.

### DIFF
--- a/scripts/.gitignore
+++ b/scripts/.gitignore
@@ -118,3 +118,7 @@ Session.vim
 *.hdf5
 *.fits
 *.parquet
+
+# Log files
+*.out
+*.log
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -75,6 +75,7 @@ REPO=/global/projecta/projectdirs/lsst/global/in2p3/Run1.2p/w_2018_30/rerun/coad
 python "${SCRIPT_DIR}"/merged_tract_cat.py ${REPO} [...]
 ```
 
+#### Run 1.2i
 Then run
 ```bash
 REPO=/global/cscratch1/sd/desc/DC2/data/Run1.2i_globus_in2p3_20181217/w_2018_39/rerun/multiband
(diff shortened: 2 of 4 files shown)


johannct:
If the question is about the scripts README, I think it can be improved indeed : 
- dependencies are not spelled out to run the suite (tables of course, which should soon be part of the stack conda)
- output products are not explicitely stated, especially the naming of the hdf5 files
- specifics of the run 1.Xy actual command lines should come after a generic exposition of the sequence of command lines to be executed.
- I iterate here a remark  I made before : it is not super pleasant to create products in-place, id est to be forced to do cd $WORKDIR in order to have the outputs be saved in $WORKDIR. I got tricked by that once, forgetting to move back outputs that had been produced in-place..... in a node scratch space that got deleted at the end of a batch job.

wmwv:
> I guess we could clean it to include only the general (and up-to-date) instructions, and create a different file to document the exact procedure used for each run/version. Is this related to what you're thinking of?

Yes, that's exactly the tension I'm concerned about.  I think this is in part why @johannct isn't happy with the current README.  It's useful guide to what was done, but it's not a full recreation of exactly what was done.  It's also very much centered around doing things at NERSC.

This isn't going to be a complete re-working.  But for this ticket I'll 
1. [ ] Add an overall introduction as a conceptual guide to what's happening.
2. [ ] List `tables` as a pretty key Python dependency that can easily not be in an install.
3. [ ] List `GCRCatalogs` and `gcr-catalogs` as a DESC-specific dependencies.

In a future ticket, @johannct can add instructions for how to run these at IN2P3.

wmwv:
@johannct 
> I iterate here a remark  I made before : it is not super pleasant to create products in-place, id est to be forced to do cd $WORKDIR in order to have the outputs be saved in $WORKDIR. I got tricked by that once, forgetting to move back outputs that had been produced in-place..... in a node scratch space that got deleted at the end of a batch job.

I think I understand the complaint, but what's your suggestion for positive action?

johannct:
ah sorry, just a --outdir option maybe?

wmwv:
> ah sorry, just a --outdir option maybe?

Sounds reasonable to me.
https://github.com/LSSTDESC/DC2-production/issues/322

PR #330, opened 2019-03-12
Issues/326 - Add DPDD Source Catalog data product

wmwv:
Source Catalog files now generated and installed at NERSC using new `merge_source_cat.py` script and associated SLURM job execution.  129,512,365 Source Ids.

Produces one Parquet file per visit.  Each file is 10-20 MB, with the entire set of 1,995 visits totaling 25 GB.  Reading these all in is slow and takes 5-15 minutes, with the variance likely due to load or memory pressure on the JupypterLab node.

Files were processed with a 8-node Taskfarmer SLURM job.  The job took 4 hours to run (preceeded 36 hours waiting in the queue)

There is an updated `scripts/README.md` that details what was done to produce these.

There is a `Notebook/verify_source_table.ipynb` to test simple properties.

There is a reader in the `issues/274` branch of `gcr-catalogs.  The above Notebook shows how to use it if you've checked out a local copy of `gcr-catalogs`.

Future work should work on performance as optimized for certain use cases.  The current performance will not scale to Run 2.1.


wmwv:
This can be run either matching to the DPDD Object Table (if passed a `--reader`) or individually reading and matching against the coadd merged detection reference catalogs via the butler (if no `--reader` is passed).  The resulting Object IDs are the same (and I have verified that they are the same).

scripts/merge_source_cat.py
```diff
@@ -0,0 +1,415 @@
...
+    # There's no separate metadata field so we redundantly include here
+    # In the Parquet storage these columns will be categoricals and not take up much disk space.
+    # Pandas automatically broadcasts a scalar to be the same shape as the DataFrame.
+    cat['visit'] = data_ref.dataId['visit']
+    cat['detector'] = data_ref.dataId['detector']
+    cat['filter'] = data_ref.dataId['filter']
+
+    # Calibrate magnitudes and fluxes
+    calib = data_ref.get('calexp_calib')
+    calib.setThrowOnNegativeFlux(False)
```
jiwoncpark: Will there be (an option for) some kind of detection cut?

scripts/merge_source_cat.py
```diff
@@ -0,0 +1,415 @@
...
+    --
+    butler:  Butler object to use to load
+    visit: int
+        Visit to process
+    filename: str
+        Filename for HDF file.
+    overwrite: bool
+        Overwrite an existing parquet file.
+    """
+    data_refs = butler.subset('src', dataId={'visit': visit})
```
jiwoncpark: Performance wise, is `butler.subset('src', dataId={'visit': visit})` followed by `butler.datasetExists()` on the data refs the same as `butler.queryMetadata(datasetType='src', format=( 'detector', 'filter'), dataId={'visit': visit}`?

scripts/merge_source_cat.py
```diff
@@ -0,0 +1,415 @@
...
+            tract_patch_data_id = {}
+            tract_patch_data_id['tract'] = tract.getId()
+            # Yes, really, the following is the supported way to get the
+            # patch ID in a way that can be queried.
+            patch_str = '%d,%d' % patch.getIndex()
+            tract_patch_data_id['patch'] = patch_str
+            if verbose:
+                print("Searching ", tract_patch_data_id)
+            try:
+                ref_table = data_ref.getButler().get(datasetType='deepCoadd_ref',
```
jiwoncpark: I see, you have to retrieve the butler object because you don't have the tract information until later. I wonder if there's something like the tracts mapping table (overlaps table) in `/global/projecta/projectdirs/lsst/global/in2p3/Run1.2i/w_2018_39/rerun/281118/tracts_mapping.sqlite3`  that allows you to select tracts and patches in advance.

scripts/merge_source_cat.py
```diff
@@ -0,0 +1,415 @@
...
+        print("Found %d objects in Object Table in region" % len(this_object_table))
+
+    if len(this_object_table) < 1:
+        # Return an array of -1
+        return np.zeros(len(cat)) - 1
+
+    this_object_table_skyCoor = SkyCoord(this_object_table['ra'],
+                                         this_object_table['dec'],
+                                         unit=(u.deg, u.deg))
+    cat_skyCoor = SkyCoord(cat['coord_ra'], cat['coord_dec'], unit=(u.rad, u.rad))
```
jiwoncpark: Just to clarify, alternatively `SkyCoord(this_object_table['coord_ra'], this_object_table['coord_dec'], unit=(u.rad, u.rad))` right?

scripts/merge_source_cat.py
```diff
@@ -0,0 +1,415 @@
...
+    else:
+        associated_ids = associate_object_ids_to_table(cat, object_table=object_table, **kwargs)
+
+    return associated_ids
+
+
+def associate_object_ids_to_coadd(cat, data_ref=None, verbose=True, **kwargs):
+    """Load and match to deepcoadd references."""
+
+    # Can I get this directly from the data_ref, or do I need data_ref.butler?
```
jiwoncpark: Resolved, correct?

Notebooks/verify_source_table.ipynb
```diff
@@ -0,0 +1,323 @@
...
+    "resid = source_table['mag'] - (-2.5*np.log10(source_table['slot_PsfFlux_instFlux']/source_table['fluxmag0']))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(resid)"
```
jiwoncpark: If this notebook is meant to be a mini tutorial later, maybe it'd be good to visualize the residuals.

Notebooks/verify_source_table.ipynb
```diff
@@ -0,0 +1,323 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Looking for objectId for one source\n",
+    "Let's look for all matching Sources for one object.\n",
+    "Let's just take the first entry from the bright_ids, assuming bright sources should be detected in the coadd and have an objectId.\n",
+    "\n",
+    "Basic checks:\n",
+    "  * Are these RA, Dec values agree?\n",
```
jiwoncpark: Sorry if I missed something! What is meant by the first question?

scripts/merge_source_cat.py
```diff
@@ -0,0 +1,415 @@
...
+    # Remove the associations that were too far away.
+    w = sep2d > matching_radius * u.arcsec
+    # sentinel value for no match.
+    # Assumes input catalog uses only non-negative numbers for object Ids
+    associated_ids[w] = -1
+
+    return associated_ids
+
+
+def prefix_columns(cat, filt, fields_to_skip=()):
```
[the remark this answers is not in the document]
wmwv: I should listen more to my linter.

scripts/merge_source_cat.py
wmwv: A SNR threshold on the Source can be done after the fact when manipulating the Source Catalog.

This here is just to prevent the wrapped C++ code from throwing an exception when taking the log of a negative number.  I'm actually planning on moving the calibration stuff out of the extraction and just recording the `fluxmag0` in future work..

scripts/merge_source_cat.py
wmwv: Not provided in the current DM Science Pipelines Gen2 Butler.

scripts/merge_source_cat.py
wmwv: If you're using a DPDD-formatted Object Table then you won't have `coord_ra` and `coord_dec`,  you will just have `ra` and `dec`.

That's why there's the creation of the `ra` and `dec` columns above in `associate_object_ids_to_coadd` where we have a coadd ref catalog from the Butler (which is `coord_ra`, `coord_dec` in `u.rad`) and convert to `ra`, `dec`, in `u.deg` and then call `associate_object_ids_to_table`.

Notebooks/verify_source_table.ipynb
wmwv: Yes, I did at first plot them.  But they're supposed to all be exactly 0.  So it was just a line of circles and matplotlib makes up some plot range and it's not clear that they're exactly 0.  I found it easier to print and get the first part and the ending part.

Notebooks/verify_source_table.ipynb
wmwv: Fixed and grammerfied to:
"Do the RA, Dec values agree across the Source measurements?"

Notebooks/verify_source_table.ipynb
wmwv: But, yes, this Notebook is clearly me convincing myself rather than thoroughly convincing anyone else.

scripts/merge_source_cat.py
wmwv: You still have to check if it exists in either case.  `queryMetadata` is just based on the registry database.  It doesn't check to make sure the product exists.

While `butler.subset` (1 ms) is about twice as slow as `butler.queryMetadata` (0.5 ms), the dominant time usage is the `datasetExists` (250 ms).

scripts/merge_source_cat.py
```diff
@@ -0,0 +1,415 @@
...
+    if args.visit_file:
+        if not args.visits:
+            args.visits = []
+
+        visits_from_file = np.loadtxt(args.visit_file, dtype=int, ndmin=1)
+        # Numpy 'int' defaults to 'numpy.int64' but then we have to
+        # Explicitly force a conversion from 'numpy.int64' to 'int'
+        # For reasons I don't fully understand,
+        # the former doesn't work with the butler.subset call
+        args.visits.extend([int(v) for v in visits_from_file])
```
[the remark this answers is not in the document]
wmwv: Hmmm... yes, that could happen.  I've added a function to ensure the uniqueness while preserving the order of the list.

wmwv:
@jiwoncpark 
Thanks for the review!  Minor comments resolved.  Small function added to ensure visit list is unique.

Could you be kind enough to take another look?

scripts/merge_source_cat.py
jiwoncpark: I see. Thank you!

scripts/merge_source_cat.py
jiwoncpark: I see, I wasn't aware of the distinction. Thank you!

wmwv:
@yymao Can you formally approve this (in your role as blessed approver) so that I can merge this.  @jiwoncpark is happy with it.

PR #338, opened 2019-05-03
Issues/327

wmwv:
Forced photometry  completed for Run 1.2p.
Adds extract scripts, including SLURM job.
Adds Notebook to inspect.

Dependent on `https://github.com/LSSTDESC/gcr-catalogs/issues/279`


wmwv:
@yymao Could you take a look at this PR?

wmwv:
Yeah, I don't have a good answer to that either.  I put it here because the verification of the catalog is part of the production.

Perhaps the distinction is that validation, which I'll use to mean roughly trying to do something scientific with the catalog, should go in `DC2-analysis`.  Probably most `DC2-production/Notebooks` for verification will seed more comprehensive Notebooks in `DC2-analysis/validation`.
I think I'll leave this here, and then once we expand to the point of validating

PR #339, opened 2019-05-07
Adds script to create merged metacal catalog

EiffL:
This PR aims to add the necessary scripts needed to produce a metacal addon catalog, accessible through the GCR, along the lines of the existing `merge_tract_cat.py` script.
I'm adding a trivial `merge_tract_metacal.py` script that should create an HDF5 file containing only the  metacal columns, which can be directly appended to the existing merged HDF5 cat.
Here is how to run this script for the currently available tract of 1.2i:
```
SCRIPT_DIR=/global/homes/w/wmwv/local/lsst/DC2-production/scripts
REPO=/global/projecta/projectdirs/lsst/global/in2p3/Run1.2i/w_2018_39/rerun/metacal-test3
TRACTS="5063"

WORKING_DIR=/global/projecta/projectdirs/lsst/global/in2p3/Run1.2i/object_catalog_new
mkdir -p "${WORKING_DIR}"
cd ${WORKING_DIR}

nohup python "${SCRIPT_DIR}"/merge_tract_metacal.py "${REPO}" ${TRACTS} > merge_tract_metacal.log 2>&1 < /dev/null
```

and this is how one would find of the metacal tracts available:
```
$ find . -name "*mcal*" | cut -c3-6 | uniq
```


EiffL:
@wmwv this allows me to export the metacal table as an HDF5 file, now I'm looking at the GCR, would you recommend using the `DC2ObjectCatalog` as a template, or a simple BaseGenericCatalog kind of like what is done for the photoz?

wmwv:
@EiffL I know this isn't in the current `merge_tract.py`, but would you be happy outputting a Parquet file instead of HDF.  We would plan to write one Parquet file per tract.

EiffL:
For sure, no problem, do you already have a script for that?

wmwv:
Not for the Object table.  You can look at `merge_source_cat.py` for writing ot Parquet, but it's not chunked by tract.

In this case, instead of putting out a file for each patch:

```
        patch_metacal_cat.to_hdf(filename, key, format='fixed')
```

```

**Comment by EiffL** [comment]
ok, that should be easy, but do you already have a GCR interface  that can read an Object table saved in parquet?

**Comment by wmwv** [comment]
Yes.  Most directly copy `GCRCatalogs/dc2_source.py`
But better would be to use what I'm doing on https://github.com/LSSTDESC/gcr-catalogs/issues/#290, where I'm inheriting from a slightly more generic `DC2Catalog` and then customizing only the quantity_modifiers.

**Comment by EiffL** [comment]
so things are a little bit complicated, right now we don't have a metacal catalog for all patches in the processed tract, and this given run doesn't have forced photometry. So implementing a composite catalog doesn't make sense yet.

Nevertheless, I have added a `merge_forced_cat.py` script that does what you wanted, i.e. directly exports a trimed merged object file as a parquet file. In addition, I have a `merge_metacal_cat.py` which only exports the aggregated metacal info in a parquet file.

Ultimately, we can use these two files to make a composite catalog, but for now I also directly append the ref catalog to the metacal catalog, because of the partially processed field.

More generally, this way of aggregating the data in two separate steps is pretty dangerous, if for some reason a metacal patch is missing, this will change the order and number of rows of part of the tract, not sure what will happen in the composite catalog reader in that case.

**Comment by wmwv** [comment]
@EiffL We should be able to match on the coadd ObjectId as an inner join to construct the composite catalog.

**Comment by EiffL** [comment]
@wmwv I'm looking at this again to make a GCR catalog of run 2.1 mcal data. I don't see any gcr catalog yet for 2.1i object catalogs, is it because the plan is to store the data in a different format? Otherwise would this script be  ok for 2.1i?

**Comment by yymao** [comment]
@EiffL It'll be in the same format. Also see https://github.com/LSSTDESC/gcr-catalogs/pull/320. We'll likely merge this next week.

**Comment by EiffL** [comment]
@yymao @johannct While you are looking  into #358 , this is the script I use to produce the merged metacal catalogs, it's essentially exactly the same thing.
To run it at nersc using shifter:
```
$ shifter --image=lsstdesc/stack-jupyter:prod
$ bash make_run2p1i_dr1b_merge_mcal_cats.sh
```

johannct:
@yymao @EiffL I can have only one branch running at a time in prod :) Can you provide the command line here?

johannct:
homogeneize naming convention : "_cat.py" versus "_catalog.py". The former, while shorter, is also unambiguous AFAICT

johannct:
keep homogeneity with DM : `--patch` instead of `--patches`.

johannct:
As the script is meant to be run on several patches separately, which are to be merged afterwards, the filename needs to capture the patches string. See for instance https://github.com/LSSTDESC/DC2-production/blob/17145e5e28abcb1fcff7d9f0e00a648378bee3c4/scripts/make_object_catalog.py#L73

johannct:
The patches are to be provided as strings, and split by "^", not as array : 
https://github.com/LSSTDESC/DC2-production/blob/17145e5e28abcb1fcff7d9f0e00a648378bee3c4/scripts/make_object_catalog.py#L203

yymao:
@EiffL we can probably merge this script with `make_object_catalog.py` by adding an option to choose between `deepCoadd_forced_src` and `deepCoadd_mcalmax_deblended`.

johannct:
@yymao  fine with me. Then the merging of patch parquet files into tract parquet files can also be done in a single script

EiffL:
@yymao , ok, so should I make the necessary modifications back on #358 ?

yymao:
@EiffL sure!

johannct:
I think the merge of the metacal script into make_object_catalog.py did not occur yet. Is it possible to move this up in the priority list? I can also try to help here.... currently my DC2-production state is not clean in the pipeline setup.

PR #353, opened 2019-06-20
Issues/337

wmwv:
Generate DIA Object table.  Compute variability statistics from DIA Source GCR catalog.

Specifically for Run 1.2p test.
Files installed at NERSC.

Verification Notebook to demonstrate that DIA Source and DIA Object are reasonable.
(and yes, an enhanced version of this notebook will be added to DC2-analysis soon.)


wmwv:
Note that the loading and processing of the DIA Object statistics is currently pessimal and takes a ridiculously long time.  This should be improved for the future, but much of that work should likely happen with `dia_pipe` in the association driver rather than in the processing here.

wmwv:
Thanks, @yymao .  I created a new section for the DIA table creation because the relationship between DIASource and DIAObject tables is fundamentally different than that between Source and Object tables.

PR #357, opened 2019-07-03
Issues/355 - Produce trim versions of Run 2.1i DR1 Object Table

wmwv:
* Produce the trim versions of the Run 2.1i DR1 ('dr1b') Object Table files from the IN2P3 Science PIpelines processing, which already generates `object_tract_????.hdf5` files.
* There were some bug fixes along the way in preparing this
* There was one somewhat confounding UnicodeDecodingError for one patch.  This patch was just skipped.
* Add documentation to README.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -173,7 +173,7 @@ python "${SCRIPT_DIR}"/trim_tract_cat.py /global/projecta/projectdirs/lsst/produ
 
 Run 2.1i dr1b
 ```bash
-python "${SCRIPT_DIR}"/trim_tract_cat.py /global/projecta/projectdirs/lsst/global/in2p3/Run2.1i/w_2019_19-v1/dpdd/calexp-v1\:coadd-dr1b-v1/object_table_summary/object_tract_*.hdf5
+python "${SCRIPT_DIR}"/trim_tract_cat.py /global/projecta/projectdirs/lsst/production/DC2_ImSim/Run2.1i/dpdd/calexp-v1\:coadd-dr1b-v1/object_table_summary/object_tract_*.hdf5
 ```
 
 ### Update gcr-catalog
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -223,9 +223,10 @@ In addition to renaming columns, this also translates to derived columns that ar
 For speed, this is done by default on the already trimmed object tables (from the step just above).  But it would be possible to do it directly from the full Object merged_tract_cat files instead.
 
 ```bash
-python "${SCRIPT_DIR}"/convert_merged_tract_to_dpdd.py --reader dc2_object_run1.1p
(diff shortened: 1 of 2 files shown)


scripts/trim_tract_cat.py
```diff
@@ -96,9 +106,11 @@ def make_trim_file(infile, output_file=None, output_dir=None,
 """)
     parser.add_argument('--output_dir', default='./',
                         help='Output directory.  (default: %(default)s))')
+    parser.add_argument('--verbose', default=False, action='store_true')
```
yymao: `store_true` by itself already means default is `False`
```suggestion
    parser.add_argument('--verbose', action='store_true')
```

scripts/trim_tract_cat.py
```diff
@@ -55,11 +56,20 @@ def make_trim_file(infile, output_file=None, output_dir=None,
...
+    with pd.HDFStore(infile, 'r', errors='replace') as fh:
         for key in fh:
+            if verbose:
+                print("Key: ", key)
             if not re.match(GROUP_PATTERN, key.lstrip('/')):
                 continue
-            load_trim_save_patch(output_file, fh, key, columns_to_keep)
+            try:
+                load_trim_save_patch(output_file, fh, key, columns_to_keep)
+            except UnicodeDecodeError as e:
```
yymao: Is `UnicodeDecodeError` actually expected and should be skipped, or is this just a hack for now and we should come up with a better solution later? If the latter case, please add a `#TODO` for future us.

scripts/README.md
```diff
@@ -161,9 +163,18 @@ This trim step uses the Generic Catalog Reader to determine the columns required
...
+```
+
+Run 2.1i dr1a
+```bash
+python "${SCRIPT_DIR}"/trim_tract_cat.py /global/projecta/projectdirs/lsst/production/DC2_ImSim/Run2.1i/dpdd/calexp-v1:coadd-v1/object_table_summary/object_tract_*.hdf5
 ```
 
+Run 2.1i dr1b
+```bash
+python "${SCRIPT_DIR}"/trim_tract_cat.py /global/projecta/projectdirs/lsst/global/in2p3/Run2.1i/w_2019_19-v1/dpdd/calexp-v1\:coadd-dr1b-v1/object_table_summary/object_tract_*.hdf5
```
yymao: should we update the path here?

scripts/trim_tract_cat.py
wmwv: It's a hack.  But I don't plan on ever actually fixing it, per se.
I plan on making sure works when we
(1) switch to do our own Parquet files; and the (2) switch to using DM Science Pipelines-produced outputs.

I've add a note to that effect in comments next to the relevant line.

scripts/trim_tract_cat.py
wmwv: Thanks, I didn't know that.

PR #358, opened 2019-07-19
add script to generate object catalog in parquet format

### LINKED ISSUE #342: Convert DC2 object catalogs into parquet format
Newer DC2 DPDD catalogs (source, forced source, dia source) are now all in parquet format with consistent schema across files, and they all use readers that are subclasses of `DC2DMCatalog`. 

On the other hand, the original DC2 DPDD object catalogs are still in hdf5 format, and hence for now we need to maintain two readers that are similar in nature. The reader of the DC2 DPDD object catalogs also has to fill in missing columns. 

We should convert the DC2 DPDD object catalogs into the same parquet format as the newer DC2 DPDD catalogs.

yymao:
This PR adds a script to generate object catalog in parquet format. This new script is called `generate_object_catalog.py` and replaces the functionality of `merge_tract_cat.py`. 

This is now a draft PR. The function `merge_coadd_forced_src` has been tested, but the CLI interface has not been tested. Also, the magnitude extraction part still needs to be fixed. 

This PR will fix #342.

### DIFF
--- a/scripts/make_object_catalog.py
+++ b/scripts/make_object_catalog.py
@@ -4,19 +4,22 @@ make_object_catalog.py
 Save catalogs to parquet from forced-photometry coadds across available filters.
 """
 import os
+import re
 
 import numpy as np
-from astropy.table import join
+from astropy.table import hstack
 
 from lsst.daf.persistence import Butler
 from lsst.daf.persistence.butlerExceptions import NoResults
 
-_default_fill_value = {'i': -1, 'b': False, 'U': ''}
 
 def _ensure_butler_instance(butler_or_repo):
     if not isinstance(butler_or_repo, Butler):
         return Butler(butler_or_repo)
     return butler_or_repo
+
+
+_default_fill_value = {'i': -1, 'b': False, 'U': ''}
 
 def _get_fill_value(name, dtype):
     kind = np.dtype(dtype).kind
@@ -55,15 +58,11 @@ def generate_object_catalog(output_dir, butler, tract, patches=None,
         butler = _ensure_butler_instance(butler)
         skymap = butler.get(datasetType='deepCoadd_skyMap')
         patches = ['%d,%d' % patch.getIndex() for patch in skymap[tract]]
-    else:
-        try:
-            patches = patches.split(',')
-        except AttributeError:
-            pass
-        else:
-            if not all(len(p) == 2 for p in patches):
-                raise ValueError('patches should be a list or a string in "11,22,33" format')
-            patches = ['{},{}'.format(*p) for p in patches]
+    elif hasattr(patches, 'split'):
+        patches = patches.split('^')
+
+    if not all(re.match(r'\d,\d$', p) for p in patches):
+        raise ValueError('patches should be a list or a string in "1,1^2,2^3,3" format')
 
     for patch in patches:
         if verbose:
@@ -98,7 +97,8 @@ def generate_object_catalog(output_dir, butler, tract, patches=None,
 
 
 def merge_coadd_forced_src(butler, tract, patch, keys_join_on=('id',),
-                           filters='ugrizy', verbose=False, return_pandas=True):
+                           filters='ugrizy', verbose=False, return_pandas=True,
+                           debug=False):
     """Load patch catalogs.  Return merged catalog across filters.
 
     butler: Butler object or str
@@ -135,14 +135,14 @@ def merge_coadd_forced_src(butler, tract, patch, keys_join_on=('id',),
             print("  No good isPrimary entries for tract {}, patch {}".format(tract, patch))
         return
 
-    merged_cat = ref_table[isPrimary]
-    del ref_table
-
-    merged_cat['tract'] = int(tract)
-    merged_cat['patch'] = str(patch)
+    ref_table = ref_table[isPrimary]
+    ref_table['tract'] = int(tract)
+    ref_table['patch'] = str(patch)
 
     cat_dtype = None
     missing_filters = list()
+    tables_to_merge = [ref_table]
+
     for filter_this in filters:
         filter_name = filters.get(filter_this) if hasattr(filters, 'get') else filter_this
         this_data_id = dict(tract_patch_data_id, filter=filter_name)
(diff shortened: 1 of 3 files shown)


yymao:
OK, this is now working and ready for review! Here's an example of how to run several patches:

```bash
python generate_object_catalog.py <REPO> 4850 --patches=00,01,02,03 --output_dir=<OUTPUT_DIR>
```

Here's a real case on NERSC:
```bash
python generate_object_catalog.py /global/cscratch1/sd/desc/DC2/data/Run2.1i/rerun/coadd-v1 4850 --patches=00,01 --output_dir=$SCRATCH --verbose
```

cc @johannct

scripts/make_object_catalog.py
```diff
@@ -0,0 +1,236 @@
...
+            cat = butler.get(datasetType='deepCoadd_forced_src',
+                             dataId=this_data_id)
+        except NoResults as e:
+            if verbose:
+                print("  ", e)
+            missing_filters.append(filter_this)
+            continue
+
+        cat = cat.asAstropy()
+        cat = cat[isPrimary]
```
yymao: @wmwv looking at the script again, I am actually confused. Isn't the fact that we can do this line (note that `isPrimary` is from `deepCoadd_ref` and used by all filters) means that we don't need to match by `id` at all? We should be able to just do a `hstack`, since the row order/number must be the same for all filters??

johannct:
Sorry @yymao to have been unclear when we chatted. The standard patch syntax is X,Y and if there are several patches, then using the stack nomenclature may be in order : in your example above 0,0^0,1^0,2^0,3 rather than 00,01,02,03

johannct:
If this goes all the way to parquet files, is it ok to run it several times with the same tract but disjoint sets of patches in order to build the full catalogue?

johannct:
[tanugi@cca001 scripts]$ git diff make_object_catalog.py
diff --git a/scripts/make_object_catalog.py b/scripts/make_object_catalog.py
index 6cb5c1c..edb02f3 100644
--- a/scripts/make_object_catalog.py
+++ b/scripts/make_object_catalog.py
@@ -57,13 +57,12 @@ def generate_object_catalog(output_dir, butler, tract, patches=None,
         patches = ['%d,%d' % patch.getIndex() for patch in skymap[tract]]
     else:
         try:
-            patches = patches.split(',')
+            patches = patches.split('^')
         except AttributeError:
             pass
         else:
-            if not all(len(p) == 2 for p in patches):
+            if not all(len(p) == 3 for p in patches):
                 raise ValueError('patches should be a list or a string in "11,22,33" format')
-            patches = ['{},{}'.format(*p) for p in patches]
 
     for patch in patches:
         if verbose:

yymao:
Thanks @johannct, I've updated the format for specifying `patches` as suggested. 

You asked:
> If this goes all the way to parquet files, is it ok to run it several times with the same tract but disjoint sets of patches in order to build the full catalogue?
Yes, each patch will have its own output, so you can run disjoint sets of patches in parallel. Once everything is done, we need to run another script to join the patches in each tract.

scripts/make_object_catalog.py
yymao: @wmwv I've updated the script to use `hstack` instead of `join` and with limited tests I believe it works as expected. If you can confirm that indeed `deepCoadd_forced_src` and `deepCoadd_ref` are expected to have the same row order/number, that would be great.

johannct:
ok then for now I can also run on tracts only, that will do away with the second script. Is there another step? I tested the script successfully by the way.

yymao:
Right, so there will be two steps:

First, generate per-patch files:
```base
REPO=/path/to/butler/repo
TRACT=4850
OUTPUTDIR=/path/to/output_dir

python make_object_catalog.py $REPO $TRACT --patches='0,0^0,1^0,2' --output-dir=$OUTPUTDIR
python make_object_catalog.py $REPO $TRACT --patches='1,0^1,1^1,2' --output-dir=$OUTPUTDIR
```
And after all patches in this tract is done, then run:

```
python merge_parquet_files.py $OUTPUTDIR/object_$TRACT_*.parquet -o=$OUTPUTDIR/object_tract_$TRACT.parquet --sort-input-files
```

EiffL:
??? Why do you have to generate per patch object catalogs ?? Michael's script  were already automatically finding and extracting the object catalogs from the patches using butler magic

yymao:
@EiffL because we may want to parallelize the patches

johannct:
keep homogeneity with DM : `--patch` instead of `--patches`.

scripts/make_object_catalog.py
```diff
@@ -0,0 +1,238 @@
...
+    return fill_value
+
+
+def generate_object_catalog(output_dir, butler, tract, patches=None,
+                            overwrite=True, verbose=False,
+                            filename_prefix='object',
+                            parquet_engine='pyarrow',
+                            **kwargs):
+    """Save catalogs to parquet from forced-photometry coadds across available filters.
+    Iterates through patches, saving each in append mode to the save HDF5 file.
```
wmwv: Update doc string.  It writes to Parquet and doesn't append.

scripts/make_object_catalog.py
```diff
@@ -0,0 +1,238 @@
...
+    butler: Butler or str (of a repo)
+        File location of Butler repository+rerun to load.
+    tract: int
+        Tract of sky region to load
+    filename_prefix: str
+        filename will be of the form "%s_%d_%s.parquet" % (filename_prefix, tract, patch)
+        With the addition that the comma will be removed from the patch name
+        to provide a valid Python identifier: e.g., 'coadd_4849_11'
+    overwrite: bool
+        Overwrite an existing HDF file.
```
wmwv: 'HDF' -> 'Parquet'

scripts/make_object_catalog.py
```diff
@@ -0,0 +1,238 @@
...
+                        help="""(default: %(default)s)""")
+    args = parser.parse_args()
+
+    filters = 'ugrizy'
+    if args.hsc:
+        filters = {'u': 'HSC-U', 'g': 'HSC-G', 'r': 'HSC-R', 'i': 'HSC-I',
+                   'z': 'HSC-Z', 'y': 'HSC-Y'}
+
+    if len(args.tract) > 1 and args.patches:
+        print("You specified more than 1 tract but only need partical patches??")
```
wmwv: "partial"

wmwv:
@johannct Because of Python's unique-prefix rules for processing options, you can still use `--patch` instead of `--patches`.

We use `--patches` in `merge_dia_object.py`.
We can change both to `--patch` in some future PR if this ends up being confusing.

PR #365, opened 2019-09-04
Adds script to generate parquet metacal catalog

EiffL:
Trying to rebase this branch to master I somehow deleted the entire conversation on adding a script to convert metacal patch catalogs to parquet :-/ 

This PR aims at adding a script equivalent to `make_object_catalog.py` but to generate metacal parquet files. @yymao suggested directly modifying `make_object_catalog.py` to optionally process the metacal catalogs, but while implementing it seemed a separate script made more sense, it would have otherwise required a lot of modifications to the object catalog scripts, which I didn't want to mess with.

Functionally, this `make_metacal_catalog.py` script can be called with the same command line as `make_object_catalog.py`, yielding similar output files, one per patch.

I'm assuming there is a subsequent merging step for the object catalog paquet files, the same thing would be needed here.

Also tagging @johannct who is the one who needs this PR.

### DIFF
--- a/scripts/make_metacal_catalog.py
+++ b/scripts/make_metacal_catalog.py
@@ -139,7 +139,10 @@ def load_metacal_patch(butler, tract, patch, verbose=False, return_pandas=True,
         assert (metacal['id'] == ref_table['id']).all()
 
     # Dropping redundant columns with the main reference catalog
-    metacal = metacal.drop(["coord_ra", "coord_dec", "parent", "id"], axis=1)
+    del metacal["coord_ra"]
+    del metacal["coord_dec"]
+    del metacal["parent"]
+    del metacal["id"]
 
     return metacal.to_pandas() if return_pandas else metacal
 
--- a/scripts/make_metacal_catalog.py
+++ b/scripts/make_metacal_catalog.py
@@ -1,0 +1,185 @@
+"""
+make_metacal_catalog.py
+
+Save catalogs to parquet from metacal runs across available patches.
+"""
+import os
+import re
+
+import numpy as np
+from astropy.table import hstack
+
+from lsst.daf.persistence import Butler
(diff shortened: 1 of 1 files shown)


scripts/make_metacal_catalog.py
```diff
@@ -0,0 +1,188 @@
...
+        return Butler(butler_or_repo)
+    return butler_or_repo
+
+
+def generate_metacal_catalog(output_dir, butler, tract, patches=None,
+                             overwrite=True, verbose=False,
+                             filename_prefix='metacal',
+                             parquet_engine='pyarrow',
+                             **kwargs):
+    """Save catalogs to parquet from individual metacal patches.
```
johannct: """Save catalogs to parquet from individual metacal tract.

scripts/make_metacal_catalog.py
```diff
@@ -0,0 +1,188 @@
...
+    Iterates through patches, saving each in append mode to the save parquet file.
```
johannct: Iterates through patches, saving each in append mode to the same parquet file.

scripts/make_metacal_catalog.py
EiffL: hum, this scripts outputs a parquet file for each (tract,patch),  they are named "metacal_1450_11.parquet" or something like that. It's the same logic as in make_object_catalog.py
Is that ok? or would you rather output a single file per tract?

scripts/make_metacal_catalog.py
johannct: You are right sorry. At some point I thought it would append all the patches provided as args. It would be more logical IMHO, but it can be discussed later

EiffL:
Is this ready to merge @johannct

johannct:
I let you delete the branch

PR #382, opened 2019-12-12
fix the call to calibration dataset _photoCalib instead of _calib, an…

johannct:
…d remove the computation of mag, magerrs and SNR in case of _photoCalib; comment out two offending lines in write_gcr_to_parquet.py that are incompatible with pyarrow 0.13

### DIFF
--- a/scripts/merge_forced_source_cat.py
+++ b/scripts/merge_forced_source_cat.py
@@ -121,8 +121,19 @@ def load_detector(data_ref, object_table=None, matching_radius=1,
 
     # Calibrate magnitudes and fluxes
     calib = data_ref.get('calexp_photoCalib')
-    # For now we just extract the grey FLUXMAG0
-    cat['FLUXMAG0'] = calib.getInstFluxAtZeroMagnitude()
+
+    mag, mag_err = calib.instFluxToMagnitude(cat[flux_names['psf_flux']].values,
+                                      cat[flux_names['psf_flux_err']].values)
+
+    cat['mag'] = mag
+    cat['mag_err'] = mag_err
+    cat['SNR'] = np.abs(cat[flux_names['psf_flux']] /
+                        cat[flux_names['psf_flux_err']])
+
+    # TODO Need to calibrate fluxes as well
+    # For now we'll just save this information in an additional column
+    # to be available further downstream.
+    cat['fluxmag0'] = calib.getInstFluxAtZeroMagnitude()
 
     # Restrict to columns that we need
     cat = cat[columns_to_keep]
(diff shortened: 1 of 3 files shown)


johannct:
Note that merge_source_cat.py had an exception handling for former `_calib` dataset, which I did not remove and into which I indented the lines that are now offending with `_photoCalib` API. We may want to homogeneize these two scripts, and we likely should give some more info in header docstring about when to use either. My understanding is that merge_source is for the Source table (table 6 in LSE-163) while merge_forced_source is for the ForcedSource table (table 7 in LSE-163) which only has 7 columns in the project document....

scripts/write_gcr_to_parquet.py
```diff
@@ -96,8 +96,8 @@ def write_dataframe_to_files(
     # Append iff the file already exists
     parquet_append = append and os.path.exists(output_filename)
     df.to_parquet(output_filename,
-                  append=parquet_append,
-                  file_scheme=parquet_scheme,
+                  #append=parquet_append,
+                  #file_scheme=parquet_scheme,
```
wmwv: Why are these lines connected to this PR?

wmwv: I think this change should be a separate, trivial PR.  Then we we want to revert it with an updated `pyarrow` it will be easy to identify the change.

scripts/merge_forced_source_cat.py
```diff
@@ -120,22 +120,9 @@ def load_detector(data_ref, object_table=None, matching_radius=1,
...
-                        cat[flux_names['psf_flux_err']])
-
-    # TODO Need to calibrate fluxes as well
-    # For now we'll just save this information in an additional column
-    # to be available further downstream.
-    flux_mag0, flux_mag0_err = calib.getFluxMag0()
-    cat['fluxmag0'] = flux_mag0
+    calib = data_ref.get('calexp_photoCalib')
+    # For now we just extract the grey FLUXMAG0
+    cat['FLUXMAG0'] = calib.getInstFluxAtZeroMagnitude()
```
wmwv: lower-case:
`cat['fluxmag0']`

wmwv: I think we should keep the 'mag', 'mag_err', and 'SNR' columns as part of this PR.  
This PR is just about fixing the _calib -> _photoCalib transition.
Removing the 'mag' columns from the catalogs (which we should do eventually) should be done in a coordinated fashion with updates to `gcr-catalogs`.

wmwv:
And my apologies that this took me 2 months to look at!

Issue #381, opened 2019-12-11
pyarrow 0.13 not compatible with write_gcr_to_parquet


johannct:
https://github.com/LSSTDESC/DC2-production/blob/d9380973d08152ba8e7948eda3d3afe0f36b4caf/scripts/write_gcr_to_parquet.py#L99-L100

These two lines cause a crash at execution : `TypeError: __cinit__() got an unexpected keyword argument 'append'` and `TypeError: __cinit__() got an unexpected keyword argument 'file_scheme'`

heather999:
And that's a problem if we need to move to some later version of pyarrow - DM is currently "stuck" at 0.13.0 so we cannot just upgrade within the stack conda env.  Either the script has to be modified to work with pyarrow 0.13.0, or we set up another conda env specifically to run `write_gcr_to_parquet.py` using whatever version of pyarrow is needed.

johannct:
I think that our scripts are too old for 0.13, not the contrary.

yymao:
`write_gcr_to_parquet.py` has been updated and this issue should be sovled.

PR #384, opened 2020-02-26
update to recent pyarrow

johannct:
change write_gcr_to_parquet.py so that 2 arguments now obsolete to drive pyarrow are removed


wmwv:
> now obsolete to drive pyarrow

I don't think I quite follow this.  What do you mean by "obsolete"?

wmwv:
Do you mean:

heather999:
@wmwv I think the issue was that our use of append was not compatible with the "newer" pyarrow 0.13.0 that DM now provides as part of its distribution:  #381

wmwv:
@johannct Do you need anything further to merge this?

PR #405, opened 2020-12-07
add healpix option in write_gcr_to_parquet.py (for converting cosmoDC2 to parquet)

yymao:
This PR adds an `--healpix` option in `scripts/write_gcr_to_parquet.py` so that it is more convenient to use this script to convert cosmoDC2 to parquet. The option can be used as:

```base
python ~/desc/DC2-production/scripts/write_gcr_to_parquet.py cosmoDC2_v1.1.4_image --healpix=8786
```

As a test, I am generating a few healpix pixels in `/global/cscratch1/sd/yymao/desc/cosmodc2-parquet` on NERSC if anyone wants to take a look. (cc @JoanneBogart @plaszczy @cwwalter)

### DIFF
--- a/scripts/write_gcr_to_parquet.py
+++ b/scripts/write_gcr_to_parquet.py
@@ -104,6 +104,8 @@ def convert_cat_to_parquet(reader,
     else:
         if is_tract_catalog and kwargs.get('tract'):
             output_filename_this = output_filename.format('_tract{}'.format(kwargs['tract']))
+        elif kwargs.get('healpix_pixels'):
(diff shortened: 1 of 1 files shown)


yymao:
On cori login node with SCRATCH file system, converting on one healpix of cosmoDC2 takes about 11 minutes, resulting in a 25 GB parquet file (per healpix, not including native quantities).

yymao:
Note of self: need to prevent duplicated columns when all column names are in lower case.

JoanneBogart (approved):
LGTM

yymao:
Thanks @JoanneBogart (and also thanks to @plaszczy and @cwwalter for checking the output files off github).

PR #409, opened 2020-12-11
Implement matching limits for truth-match catalog

yymao:
This PR implements the truth matching limits (dmag < 1 and sep < 1") as described in the DC2 paper. For objects that do that have match with these limits, nearest neighbor is used as a match. A `is_good_match` column indicates whether the matching limits are met. 

With this implementation, the majority of matching entries are identical with Javi's catalog, but there's 0.1% of matches that differ. Upon inspection, this difference is due to the different sources of magnitude --- I used the magnitude in the truth catalog, while Javi used cosmoDC2. Because the algorithm choose the closest magnitude within 1" when multiple candidates exist, the small difference in magnitude results in different matches in the faint end.

image

Note that this PR is current made w.r.t to the `u/yymao/allow-multiple-files` branch to highlight the relevant change. But it should be merged in to master in the end.

### DIFF
--- a/scripts/merge_truth_per_tract.py
+++ b/scripts/merge_truth_per_tract.py
@@ -9,10 +9,12 @@ import re
 import warnings
 import multiprocessing as mp
 from argparse import ArgumentParser, RawTextHelpFormatter
+from collections import defaultdict
 
 import numpy as np
-from astropy.coordinates import SkyCoord
 import pandas as pd
+import astropy.units as u
+from astropy.coordinates import SkyCoord, search_around_sky
 
 __all__ = ["merge_truth_per_tract", "match_object_with_merged_truth"]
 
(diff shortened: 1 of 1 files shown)


scripts/merge_truth_per_tract.py
```diff
@@ -82,60 +83,124 @@ def merge_truth_per_tract(input_dir, truth_types=("truth_", "star_", "sn_"), val
...
-        "match_objectId": object_cat["objectId"]
+        "dmag": _calc_dmag(object_idx, truth_idx),
     })
-    del idx, sep
-
-    matched.sort_values("match_sep", inplace=True)
+    del object_idx, truth_idx, sep
+
+    matched = matched.sort_values("match_sep")
+    matched["is_nearest_neighbor"] = ~matched.duplicated("object_idx", keep="first")
```
fjaviersanchez, line 163: neat!

fjaviersanchez (approved):
Thanks a lot @yymao for implementing this. The code looks good and the results look reasonable to me.

yymao:
Thank you @fjaviersanchez! 

I'll merge this after #407 is merged.

PR #411, opened 2020-12-14
Ensure schema is consistent across files generated by write_gcr_to_parquet.py

yymao:
This PR updates `write_gcr_to_parquet.py` to ensure the parquet schema is consistent across files  generated within one GCR catalog.

The issue is first noticed by @wmwv in https://github.com/LSSTDESC/gcr-catalogs/pull/515#issuecomment-744687756. The cause is that `get_quantites` returns a python dictionary. While a python dictionary is always ordered for Python 3.7+, the implementation of `get_quantites` does not make such assumption nor an effort to keep it ordered. Hence, the returned dictionary can have keys in different orders every time.

This PR made two changes to ensure schema consistency:
1. First, in `_chunk_data_generator` where an arrow table is created from the returned dictionary of `get_quantites`, we now iterate over sorted columns instead of dictionary keys. 
2. `_write_one_parquet_file` now accepts a schema input as an optional argument, and hence multiple calls to `_write_one_parquet_file` can share the same schema, ensure the resulting files must have the same schema. 

Note: this PR is made with respect to the branch in #407, so that the new changes can be more easily identified. This PR should be merged into master after #407 is merged.

### DIFF
--- a/scripts/write_gcr_to_parquet.py
+++ b/scripts/write_gcr_to_parquet.py
@@ -63,7 +63,7 @@ class Checkpoint():
 def _chunk_data_generator(cat, columns, native_filters=None):
     columns = sorted(columns)
     for data in cat.get_quantities(columns, native_filters=native_filters, return_iterator=True):
-        table = pa.Table.from_arrays(*zip(*[(col, pa.array(data[col])) for col in columns]))
+        table = pa.Table.from_arrays([pa.array(data[col]) for col in columns], columns)
         del data
         try:
             cat.close_all_file_handles()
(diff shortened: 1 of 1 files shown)


wmwv (approved):
LGTM

I haven't run.  But the logic and the functionality make sense.

yymao:
Thanks @wmwv -- Files from this can be found on NERSC at

    /global/cscratch1/sd/yymao/desc/dc2_object_run2.2i_dr6_wfd_v1.new2

yymao:
The following code snippet now works on this new data set:

```python
import pyarrow.parquet as pq
dir_path = "/global/cscratch1/sd/yymao/desc/dc2_object_run2.2i_dr6_wfd_v1.new2"
pq.ParquetDataset(dir_path)
```

PR #415, opened 2021-01-05
Add scripts/READMD_write_gcr_to_parquet.md

yymao:
@heather999 has asked me to update `scripts/README.md` but the content is somewhat outdated and the file also seems to more like a log than instructions. 

I hence created a new file `scripts/READMD_write_gcr_to_parquet.md` to document how to use `write_gcr_to_parquet.py`. The procedures listed in this new file are what I used to generate the public release DPDD catalog.

### DIFF
--- a/scripts/README_write_gcr_to_parquet.md
+++ b/scripts/README_write_gcr_to_parquet.md
@@ -1,0 +1,98 @@
+# Using write_gcr_to_parquet.py
+
+This README contains the updated instructions for using `write_gcr_to_parquet.py`.
+It is last updated by Yao-Yuan Mao on Jan 5th, 2021.
+
+## General environment setup
+
+### Step 1: Create a new working directory in scratch
+
+Most likely you would want to put the parquet files in scratch area first,
+and then copy them to the actual destination when everything is done.
+
+```bash
+cd $SCRATCH
+mkdir -p desc/a-new-working-dir
+cd desc/a-new-working-dir
+```
+
+### Step 2: Clone the needed repos
+
+To make sure that you are using the up-to-date scripts and catalog configs,
+you should clone the repos.
+If you want to test a specific branch, change the following accordingly.
+
+```bash
+git clone --depth=1 git@github.com:LSSTDESC/gcr-catalogs.git
+git clone --depth=1 git@github.com:LSSTDESC/DC2-production.git
+```
+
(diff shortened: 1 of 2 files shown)


scripts/README_write_gcr_to_parquet.md
```diff
@@ -0,0 +1,98 @@
...
+
+```bash
+salloc -N 1 -C haswell -q interactive -A m1727 -t 04:00:00
+```
+
+### Step 5: Generate Parquet files
+
+Generally speaking you can just run:
+
+```bash
```
JoanneBogart, line 47: Saying "Generally speaking" and then giving a specific example with a real catalog can be confusing.  I suggest  replacing `cosmoDC2_v1.1.4` with a placeholder like `some-catalog`.   Or else change the text "generally speaking" to something that makes clear this is just an example.

Also since you are about to discuss changing the output filename prefix I would mention in this example what the output would be (by default, without `--output-filename` option)

scripts/README_write_gcr_to_parquet.md
```diff
@@ -0,0 +1,98 @@
+# Using write_gcr_to_parquet.py
+
+This README contains the updated instructions for using `write_gcr_to_parquet.py`.
+It is last updated by Yao-Yuan Mao on Jan 5th, 2021.
```
JoanneBogart, line 4:
```suggestion
It was last updated by Yao-Yuan Mao on Jan 5th, 2021.
```

scripts/README_write_gcr_to_parquet.md
```diff
@@ -0,0 +1,98 @@
...
+
+## General environment setup
+
+### Step 1: Create a new working directory in scratch
+
+Most likely you would want to put the parquet files in scratch area first,
```
JoanneBogart, line 10:
```suggestion
Most likely you will want to put the parquet files in a scratch area first,
```

JoanneBogart (commented):
I made some minor comments, but basically the instructions are clear.  I need to try it myself once.
However the instructions appear to only cover the case of creating the parquet files for the whole catalog. The case which just came up involved creating the file for only one tract.   This is likely to come up again.   Can you add something about how to run the script for a single tract or small number of identified tracts (or other partition element)?

JoanneBogart (approved):
LGTM. Thanks, @yymao

PR #416, opened 2021-01-15
README fix: --output-filename needs extension

yymao:
In `README_write_gcr_to_parquet.md` that I created earlier, I said that one can do `--output-filename=<prefix>` but it really should be `--output-filename=<prefix>.parquet` (unless one doesn't want the "parquet" extensions, of course). 

This PR fixes it.


JoanneBogart (approved):
LGTM

yymao:
Thanks!

PR #424, opened 2021-02-19
Add doc for converting from sqlite to other formats and draft noteboo…

### LINKED ISSUE #423: Create PostgreSQL tables and parquet files for Run3.1i truth
Truth tables in the form of sqlite files have been created for agn, lensed agn, lensed sne, and lensed host galaxies. They should be ingested into PostgreSQL and also translated to parquet.

JoanneBogart:
fix #423

### DIFF
--- a/Documents/conversions/convert_sqlite.md
+++ b/Documents/conversions/convert_sqlite.md
@@ -21,7 +21,7 @@ For more information about psql see [https://www.postgresql.org/docs/9.6/app-psq
 
 ### Creating the PostgreSQL schema
 
-From the SQLite table schema make an equivalent .sql file, say called `my_create.sql`, to create the PostgreSQL table. (You can see an SQLite schema from within the SQLite command line program by typing
+From the SQLite table schema make an equivalent .sql file, say called `my_create.sql`, to create the PostgreSQL table. You can see an SQLite schema from within the SQLite command line program by typing
 ```
 sqlite> .schema
 ```
@@ -71,11 +71,11 @@ where `list-of-column-names` should be those names in the SQLite schema for the 
 
 ### Update
 For tables with ra, dec columns (for truth tables, only truth_summary) update any values of type `earth` (can be computed from ra, dec) with an sql command something like
-```
+```sql
 update schema-name.table-name set coord=public.radec_to_coord(ra-col, dec-col);
 ```
 e.g. for star truth
-```
+```sql
 update star_truth.truth_summary set coord=public.radec_to_coord(ra, dec);
 ```
 ### Create indices
@@ -83,7 +83,8 @@ update star_truth.truth_summary set coord=public.radec_to_coord(ra, dec);
 
 ### Update permissions
 Finally, for all schemas, all tables, give read permission to user desc_dc2_drp_user. For a schema named `the_schema` this can be done from psql with these two commands:
-```
+
+```sql
 desc_dc2_drp=> grant usage on schema the_schema to desc_dc2_drp_user;
 desc_dc2_drp=> grant select on all tables in schema the_schema to desc_dc2_drp_user;
 ```
@@ -123,11 +124,13 @@ optional arguments:
 A typical invocation looks like:
 
 ```
-python parquet_utils.py agn_truth_cat.db --pqfile agn_truth_summary.parquet --table truth_summary
+$ python parquet_utils.py agn_truth_cat.db --pqfile agn_truth_summary.parquet --table truth_summary
 ```
 
 **Note:** That the repo `sims_TruthCatalog` requires the stack, but the script does not.  You can copy the files `parquet_utils.py` and `script_utils.py` from that repo to another area, modify the statement
-```from desc.sims_truthcatalog.script_utils import print_callinfo
+
+```python
+from desc.sims_truthcatalog.script_utils import print_callinfo
 ```
 in the first file appropriately, and then run it using any of the standard desc kernels.
 
--- a/Documents/conversions/convert_sqlite.md
+++ b/Documents/conversions/convert_sqlite.md
@@ -53,7 +53,7 @@ e.g. for star truth
 update star_truth.truth_summary set coord=public.radec_to_coord(ra, dec);
 ```
 ### Create indices
-`id` is a unique index for all but variability tables. Any columns of type `earth should also be indexed. For variability tables, index at least `id` and `obsHistID`. (**NOTE:** PostgreSQL ignores case so the column may appear as `obshistid`)
+`id` is a unique index for all but variability tables. Any columns of type `earth` should also be indexed. For variability tables, index at least `id` and `obsHistID`. (**NOTE:** PostgreSQL ignores case so the column may appear as `obshistid`)
 
 ### Update permissions
 Finally, for all schemas, all tables give read permission to user desc_dc2_drp_user
--- a/Documents/conversions/convert_sqlite.md
+++ b/Documents/conversions/convert_sqlite.md
@@ -84,7 +84,7 @@ update star_truth.truth_summary set coord=public.radec_to_coord(ra, dec);
 ### Update permissions
 Finally, for all schemas, all tables, give read permission to user desc_dc2_drp_user. For a schema named `the_schema` this can be done from psql with these two commands:
 
-```sql
+```
 desc_dc2_drp=> grant usage on schema the_schema to desc_dc2_drp_user;
 desc_dc2_drp=> grant select on all tables in schema the_schema to desc_dc2_drp_user;
 ```
--- a/Documents/conversions/convert_sqlite.md
+++ b/Documents/conversions/convert_sqlite.md
@@ -7,9 +7,35 @@
 * PostgreSQL schema name
 * PostgreSQL table name (defaults to sqlite tablename)
 
+### PostgreSQL prerequisites
+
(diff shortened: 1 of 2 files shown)


JoanneBogart:
Still to be resolved:
* should the module parquet_utils, currently in repo sims_TruthCatalog, be moved here? It creates a parquet file from SQLite and can do some rudimentary checks.  The problem with leaving it in sims_TruthCatalog is that the repo is dependent on the stack, which otherwise would not be necessary to run the module-level script.
* What belongs in the verification notebook? So far comparing counts and plotting coverage (for truth summary tables) are implemented. Other possibilities include comparing light curves (partially done) for a particular object, comparing histograms of column values, comparing values in rows.

Documents/conversions/convert_sqlite.md
```diff
@@ -0,0 +1,108 @@
...
+For tables with ra, dec columns (for truth tables, only truth_summary) update any values of type `earth` with an sql command something like
+```
+update schema-name.table-name set coord=public.radec_to_coord(ra-col, dec-col);
+```
+e.g. for star truth
+```
+update star_truth.truth_summary set coord=public.radec_to_coord(ra, dec);
+```
+### Create indices
+`id` is a unique index for all but variability tables. Any columns of type `earth should also be indexed. For variability tables, index at least `id` and `obsHistID`. (**NOTE:** PostgreSQL ignores case so the column may appear as `obshistid`)
```
yymao, line 56:
```suggestion
`id` is a unique index for all but variability tables. Any columns of type `earth` should also be indexed. For variability tables, index at least `id` and `obsHistID`. (**NOTE:** PostgreSQL ignores case so the column may appear as `obshistid`)
```

yymao:
I don't have strong opinion regarding the two points you raised:

- As long as the location of `parquet_utils` is well documented that's oK. If we keep it in `sims_TruthCatalog`, maybe you can add a note in this instruction saying one can run this script without stack.
- Your proposal of contents in the verification notebook sounds reasonable. If there's any validation checks that can be done automatically (i.e., without the need of human inspection), those can be written as a script or even as part of the generation script (e.g., with a `--validate` option to turn it on). 

I took a quick look of `convert_sqlite.md`. Most of it looks good. 

However, I can't quite follow the "Creating the PostgreSQL schema" section though. For example, `my_create.sql` is not used in the following steps. I don't know what `psql` is and how to run it. Is `desc_dc2_drp` a table name? The section mentions "If original table has ra,dec columns, add a column of type `earth`"; how does one do so?

JoanneBogart:
Thanks @yymao !    I wrote up the notes mostly for my own purposes so they do assume familiarity with various things, but it would be better if they were more self-contained.  I will aim for that in the next round of updates.

JoanneBogart:
I've updated the documentation to be less cryptic.   The notebook includes a direct check for ra, dec values in the summary tables in addition to plots.  Those checks and others like them should at some point be moved into a script.

JoanneBogart:
@yymao There is no rush, but please have a look when you get the chance.  I'm not planning to do any more work on this (other than fixing issues you find) in the near future and would like to merge what's here.  Thanks!

Documents/conversions/convert_sqlite.md
```diff
@@ -0,0 +1,145 @@
...
+## From SQLite to PostgreSQL
+### Inputs
+* path to sqlite file
+* table name
+* PostgreSQL schema name
+* PostgreSQL table name (defaults to sqlite tablename)
+
+### PostgreSQL prerequisites
+
+The database is called `desc_dc2_drp`.  In order to make changes to the database you must have access to the account `desc_dc2_drp_admin`.  By default PostgreSQL applications like the command-line tool psql will look for account information  in the file `.pgpass` in your home directory, readable (and writable) only by you.  Then you can invoke psql as follows:
```
yymao, line 12: I wonder if you want to leave a pointer as to how to access PostgreSQL in DESC, maybe link to https://confluence.slac.stanford.edu/display/LSSTDESC/Getting+Started+with+PostgreSQL+at+NERSC ?

Documents/conversions/convert_sqlite.md
```diff
@@ -0,0 +1,145 @@
...
+|----------------|------------------|
+| TEXT           | TEXT             |
+| INTEGER or INT | INTEGER          |
+| BIGINT         | BIGINT           |
+| FLOAT          | REAL             |
+| DOUBLE         | DOUBLE PRECISION |
+
+If the original table has ra,dec columns, add a column of type `earth` (a typical name for such a column is `coord` or something containing that string) to the PostgreSQL schema. All of this could perhaps be automated but if so, review output.
+
+Run it from psql:
```
yymao, line 40: By "run it" does it mean to create a schema in PostgreSQL according to `my_create.sql`? It's not immediately clear to me.

Documents/conversions/convert_sqlite.md
```diff
@@ -0,0 +1,145 @@
...
+
+| SQLite         | PostgreSQL       |
+|----------------|------------------|
+| TEXT           | TEXT             |
+| INTEGER or INT | INTEGER          |
+| BIGINT         | BIGINT           |
+| FLOAT          | REAL             |
+| DOUBLE         | DOUBLE PRECISION |
+
+If the original table has ra,dec columns, add a column of type `earth` (a typical name for such a column is `coord` or something containing that string) to the PostgreSQL schema. All of this could perhaps be automated but if so, review output.
```
yymao, line 38: If I want to add a column of type `earth`, do I edit `my_create.sql` with a text editor directly?

Notebooks/verify_truth.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "For each data source et delta_flux and time values for the plot and some summary information about the star. Order conveniently for plotting."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#id = df_lengths['id'][0]\n",
```
yymao, line 531: This cell is completely commented out. Do you intend to leave it in there?

Notebooks/verify_truth.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "* histogramming a column or two"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Compare counts\n",
```
yymao, line 267: This cell takes quite a while to run. Add a warning for the user?

Notebooks/verify_truth.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "sq_df.shape"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#Parquet\n",
```
yymao, line 594: This cell takes quite a while to run. Add a warning for the user?

yymao (changes requested):
@JoanneBogart sorry for the delay. I've reviewed the readme and the notebook. I left some minor comments inline.

Documents/conversions/convert_sqlite.md
JoanneBogart: Yes.  In fact, I'm assuming that, for now, "make an equivalent .sql file" is most likely done with a text editor.  I'll add something about it at that point.

Documents/conversions/convert_sqlite.md
JoanneBogart: That line `\i my_create.sql;`  has the same effect as executing each of the statements in my_create.sql.   That file will certainly have a `create table` statement.   It likely will also have a line like
`create schema if not exists some-schema-name;` (unfortunately involving a different meaning for the word "schema"; that should be explained)
I don't think it's appropriate to turn this readme into an sql tutorial, but I could include an example of an actual file used to create a table.

Notebooks/verify_truth.ipynb
JoanneBogart: Good idea.

Notebooks/verify_truth.ipynb
JoanneBogart: No, I just forgot about it. It should go.

Notebooks/verify_truth.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for band in ('i', 'r', 'g', 'u'):\n",
+    "    title = format_title(id, lc_ra, lc_dec, band, object='agn')\n",
+    "    plot_object(title, [pg_df, sq_df, pq_df], ('PostgreSQL', 'SQLite', 'Parquet'), band)"
+   ]
+  },
```
yymao, line 677:
```suggestion
```

I think this cell should be removed? Since `lc_ra`, `lc_dec` etc are no longer defined in the notebook?

yymao (changes requested):
Thanks @JoanneBogart. The changes look good to me. 

The very last cell in the validation notebook doesn't work because the variables are no longer defined. Should we remove that cell?

Notebooks/verify_truth.ipynb
JoanneBogart: I do intend to make those plots.  The mistake was to not save the values I needed from an earlier query. That's fixed now.  I also found some other minor things to clean up.

yymao (approved):
Thanks for the fix. I've verified that the notebook runs without problem. Approving now...

JoanneBogart:
Thanks, @yymao  :-)

PR #430, opened 2022-05-15
update/define columns when merging truth catalogs of different types

yymao:
This PR allows the merge of truth catalogs of different types when they have different columns. 

This is already used for the production of the data release v4 truth match catalog.


JoanneBogart (approved):
LGTM

PR #431, opened 2022-06-17
Truth match: switch to point-source mag for point sources and when cModel is not finite

yymao:
As discussed in [this Slack thread](https://lsstc.slack.com/archives/C77DDKZHR/p1655483129895869?thread_ts=1655298064.897649&cid=C77DDKZHR), currently we use cModel mag from the object catalog to calculate magnitude difference when we generate the truth match catalog. 

This PR updates this process, and switches to point-source magnitudes if 
1. the corresponding truth entry is not a galaxy; OR
2. when cModel is not finite

### DIFF
--- a/scripts/merge_truth_per_tract.py
+++ b/scripts/merge_truth_per_tract.py
@@ -161,18 +161,19 @@ def match_object_with_merged_truth(
     object_idx, truth_idx, sep, _ = search_around_sky(object_sc, truth_sc, sep_limit_arcsec * u.arcsec)  # pylint: disable=no-member
 
     # Calculate magnitude difference for all the pairs
+    # Use cModel magnitude if it's finite and the corresponding truth entry is a galaxy; otherwise use PSF magnitude
+    mag_to_match = np.where(
+        (
+            (truth_cat.loc[truth_idx, "truth_type"].values == 1)  # truth_type == 1 for galaxies
+            & np.isfinite(object_cat.loc[object_idx, mag_label_obj_cm].values)
+        ),
+        object_cat.loc[object_idx, mag_label_obj_cm].values,  # cModel mag
+        object_cat.loc[object_idx, mag_label_obj_ps].values,  # point-source mag
+    )
     with np.errstate(invalid="ignore"):
-        dmag = np.abs(
-            np.where(
-                (
-                    (truth_cat.loc[truth_idx, "truth_type"].values == 1)
-                    & np.isfinite(object_cat.loc[object_idx, mag_label_obj_cm].values)
-                ),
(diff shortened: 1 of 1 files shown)


fjaviersanchez (approved):
The fix looks good to me. Thanks!

yymao:
@wmwv No. There do exist object entries who have non-finite cmodel mag but their nearest truth entry is galaxy type. This mostly happen for faint objects, so it's not related to the originally thread.

Among these objects (i.e., having non-finite cmodel mag but their nearest truth entry is galaxy type), about half of them don't have finite point-source mag either. The remaining has valid point-source mag that can be used to matching.

scripts/merge_truth_per_tract.py
```diff
@@ -162,8 +163,15 @@ def match_object_with_merged_truth(
...
     with np.errstate(invalid="ignore"):
         dmag = np.abs(
-            object_cat.loc[object_idx, mag_label_obj].values -
-            truth_cat.loc[truth_idx, mag_label_truth].values
+            np.where(
+                (
+                    (truth_cat.loc[truth_idx, "truth_type"].values == 1)
+                    & np.isfinite(object_cat.loc[object_idx, mag_label_obj_cm].values)
+                ),
+                object_cat.loc[object_idx, mag_label_obj_cm].values,
```
wmwv, line 171: Add a comment explaining what this selection is doing here.  Seeing the `_cm` vs `_ps` can be a little hard in all of the other code.  "Match based on cModel magnitudes for true galaxy with finite measured cModel magnitudes.  Otherwise match on PSF magnitude."

scripts/merge_truth_per_tract.py
```diff
@@ -162,8 +163,15 @@ def match_object_with_merged_truth(
     # Calculate magnitude difference for all the pairs
     with np.errstate(invalid="ignore"):
         dmag = np.abs(
-            object_cat.loc[object_idx, mag_label_obj].values -
-            truth_cat.loc[truth_idx, mag_label_truth].values
+            np.where(
```
wmwv, line 166: My preference would be to split this into two lines to make it easier to document what the first quantity is
```
mag_to_match = np.where( ... )
dmag = mag_to_match - truth_cat.loc[truth_idx, mag_label_truth].values
del mag_to_match
```

It don't think that there's a memory performance difference here in terms of temporary arrays.  I think the `np.where` command is going to generate the same temporary array anyway.

wmwv (changes requested):
A suggestion to improve future readability of code by adding an explanatory comment and re-structuring into two lines.

wmwv (approved):
LGTM!

